PR #1, opened 2015-04-10
Probable typo of "tentant" instead of "tenant". Fix for pull request.

philpraxis:
For more information, contact committer and/or see NFV-SDN tasks and
work on http://labs.p1sec.com

### DIFF
--- a/opflexagent/gbp_ovs_agent.py
+++ b/opflexagent/gbp_ovs_agent.py
@@ -193,7 +193,7 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
         if device_owner == n_constants.DEVICE_OWNER_DHCP:
             ips.append(METADATA_DEFAULT_IP)
         mapping_dict = {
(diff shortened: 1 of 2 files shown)


ivar-lazzaro:
Although this is definitively a typo, we can't change it unless we fix the GBP counterpart[0]. 

[0] https://github.com/noironetworks/group-based-policy/blob/noiro/gbpservice/neutron/services/grouppolicy/drivers/cisco/apic/apic_mapping.py#L326

philpraxis:
Ok, done, see:
https://github.com/noironetworks/group-based-policy/pull/1

ivar-lazzaro:
I'll merge this in master and leave the noiro branch as is. The reason is that we don't plan on supporting the noiro branch after the features are merged on GBP upstream, therefore [0] doesn't need to be merged.

[0] https://github.com/noironetworks/group-based-policy/pull/1

PR #6, opened 2015-09-19
Updated agent dependencies to include agent-ovs

mandeepdhami:


rpm/neutron-opflex-agent.service
```diff
@@ -1,6 +1,8 @@
 [Unit]
 Description=Neutron OpFlex Plugin Agent
-After=syslog.target network.target
+After=syslog.target network.target agent-ovs.service
```
amittbose: Do we need something similar for debian? The file to update is debian/neutron-opflex-agent.upstart

mandeepdhami: Yes, I will submit a pull request for that tonight. I did not have an Ubuntu system to test on yet.

mandeepdhami:
I forgot to create a pull request for Juno branch. Will do that tonight as well.

PR #9, opened 2015-10-21
Initial checkin for opflex-notify-agent

mandeepdhami:


amittbose:
Do we really need another agent? Seems like more operational overhead. Why not just launch a new thread from gbp_ovs_agent:main to monitor the notification socket?

mandeepdhami:
Good idea. I will do that.

mandeepdhami:
Updated

PR #10, opened 2015-10-22
create multiple EP files per MAC address

ivar-lazzaro:
allow_address_pairs extension in Neutron allows multiple MAC
addresses to exist for the same port. Create one EP file for each
MAC as per the opflex afgent specifications.

### DIFF
--- a/opflexagent/gbp_ovs_agent.py
+++ b/opflexagent/gbp_ovs_agent.py
@@ -400,7 +400,10 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
                 virtual_ips.append(
                     {'ip': aap['ip_address'],
                      'mac': aap.get('mac_address', mac)})
-        mapping_dict['ip'] = ips + ips_ext
+                if aap.get('active'):
+                    ips_ext.append(aap['ip_address'])
+        if ips or ips_ext:
+            mapping_dict['ip'] = sorted(ips + ips_ext)
         if virtual_ips:
             mapping_dict['virtual-ip'] = sorted(virtual_ips,
                                                 key=lambda x: x['ip'])
--- a/opflexagent/gbp_ovs_agent.py
+++ b/opflexagent/gbp_ovs_agent.py
@@ -35,7 +35,6 @@ from oslo_serialization import jsonutils
 from opflexagent import as_metadata_manager
 from opflexagent import constants as ofcst
 from opflexagent import rpc
-from opflexagent import snat_iptables_manager
 
 eventlet_utils.monkey_patch()
 LOG = logging.getLogger(__name__)
@@ -274,15 +273,60 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
                     port, net_uuid, network_type, physical_network,
                     segmentation_id, fixed_ips, device_owner, ovs_restarted)
         elif network_type in self.supported_pt_network_types:
+            # PT cleanup is needed before writing the new endpoint files
+            self.mapping_cleanup(port.vif_id, cleanup_vrf=False)
             if ((self.opflex_networks is None) or
                     (physical_network in self.opflex_networks)):
                 # Port has to be untagged due to a opflex agent requirement
                 self.int_br.clear_db_attribute("Port", port.port_name, "tag")
-                self.mapping_to_file(port, net_uuid, mapping, fixed_ips,
+                # Multiple files will be created based on how many MAC
+                # addresses are owned by the specific port.
+                mapping_copy = copy.deepcopy(mapping)
+                mac_active_aap = {}
+                mapping_copy['allowed_address_pairs'] = []
+                mapping_copy['floating_ip'] = []
+                fip_by_fixed = {}
+                # Prepare FIPs by fixed_ip
+                for fip in mapping.get('floating_ip', []):
+                    fip_by_fixed.setdefault(
+                        fip['fixed_ip_address'], []).append(fip)
+                for aap in mapping.get('allowed_address_pairs', []):
+                    if aap.get('active'):
+                        if not aap.get('mac_address'):
+                            # Should go with the MAIN mac address EP file
+                            mapping_copy['allowed_address_pairs'].append(aap)
+                            # Also set the right floating IPs
+                            mapping_copy['floating_ip'].extend(
+                                fip_by_fixed.get(aap['ip_address'], []))
+                        else:
+                            # Store for future processing
+                            mac_active_aap.setdefault(
+                                aap['mac_address'], []).append(aap)
+                # Create mapping file for base MAC address
+                self.mapping_to_file(port, net_uuid, mapping_copy, fixed_ips,
                                      device_owner)
+                # Reset for AAP EP files
+                mapping_copy['allowed_address_pairs'] = []
+                mapping_copy['fixed_ips'] = []
+                mapping_copy['extra_ips'] = []
+                mapping_copy['subnets'] = []
+                mapping_copy['enable_dhcp_optimization'] = False
+                mapping_copy['enable_metadata_optimization'] = False
+                # Map to file based on the active AAP with a MAC address
+                for mac, aaps in mac_active_aap:
+                    # Replace the MAC address with the new one
+                    mapping_copy['mac_address'] = mac
+                    # Extend the FIP list based on the allowed IPs
+                    mapping_copy['floating_ip'] = []
+                    mapping_copy['ip-address-mapping'] = []
+                    for aap in aaps:
+                        mapping_copy['floating_ip'].extend(fip_by_fixed.get(
+                            aap['ip_address'], []))
+                    # For this mac, set all the allowed address pairs.
+                    mapping_copy['allowed_address_pairs'] = aaps
+                    self.mapping_to_file(port, net_uuid, mapping_copy, None,
+                                         device_owner)
             else:
-                # PT cleanup may be needed
-                self.mapping_cleanup(port.vif_id)
                 LOG.error(_("Cannot provision OPFLEX network for "
                             "net-id=%(net_uuid)s - no bridge for "
                             "physical_network %(physical_network)s"),
@@ -312,18 +356,24 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
         if device_owner in [n_constants.DEVICE_OWNER_ROUTER_INTF]:
             return
         ips_ext = mapping.get('extra_ips') or []
-
+        mac = mapping.get('mac_address') or port.vif_mac
         mapping_dict = {
             "policy-space-name": mapping['ptg_tenant'],
             "endpoint-group-name": (mapping['app_profile_name'] + "|" +
                                     mapping['endpoint_group_name']),
             "interface-name": port.port_name,
-            "mac": mapping.get('mac_address') or port.vif_mac,
+            "mac": mac,
             "promiscuous-mode": mapping.get('promiscuous_mode') or False,
             "uuid": port.vif_id,
             'neutron-network': net_uuid}
 
-        ips = [x['ip_address'] for x in fixed_ips]
+        # Allocated IPs should be filtered by ownership.
+        if 'owned_addresses' in mapping:
+            ips = [x['ip_address'] for x in fixed_ips if
+                   x['ip_address'] in mapping['owned_addresses']]
+        else:
+            ips = [x['ip_address'] for x in fixed_ips]
+
         virtual_ips = []
         if device_owner == n_constants.DEVICE_OWNER_DHCP:
             # vm-name, if specified in mappings, will override this
@@ -354,7 +404,8 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
             mapping_dict['domain-name'] = mapping['vrf_name']
 
         self._fill_ip_mapping_info(port.vif_id, mapping, ips, mapping_dict)
-        self._write_endpoint_file(port.vif_id, mapping_dict)
+        # Create one file per MAC address.
+        self._write_endpoint_file(port.vif_id + '_' + mac, mapping_dict)
         self.vrf_info_to_file(mapping, vif_id=port.vif_id)
         if mapping.get('enable_metadata_optimization', False):
             self.metadata_mgr.ensure_initialized()
@@ -410,16 +461,16 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
             mapping_dict['dhcp6'] = {
                 'dns-servers': v6subnets[0]['dns_nameservers']}
 
-    def mapping_cleanup(self, vif_id):
+    def mapping_cleanup(self, vif_id, cleanup_vrf=True):
         LOG.debug('Cleaning mapping for vif id %s', vif_id)
-        self._delete_endpoint_file(vif_id)
+        self._delete_endpoint_files(vif_id)
         es = self._get_es_for_port(vif_id)
         self._dissociate_port_from_es(vif_id, es)
         self._release_int_fip(4, vif_id)
         self._release_int_fip(6, vif_id)
         vrf_id = self.vif_to_vrf.get(vif_id)
         vrf = self.vrf_dict.get(vrf_id)
-        if vrf:
+        if vrf and cleanup_vrf:
             del self.vif_to_vrf[vif_id]
             vrf['vifs'].discard(vif_id)
             if not vrf['vifs']:
@@ -515,6 +566,17 @@ class GBPOvsAgent(ovs.OVSNeutronAgent):
 
     def _delete_endpoint_file(self, port_id):
         return self._delete_file(port_id, self.epg_mapping_file)
+
+    def _delete_endpoint_files(self, port_id):
+        # Delete all files for this specific port_id
+        directory = os.path.dirname(self.epg_mapping_file)
+        # Remove all existing EPs mapping for port_id
+        for f in os.listdir(directory):
+            if f.endswith('.' + FILE_EXTENSION) and port_id in f:
+                try:
+                    os.remove(os.path.join(directory, f))
+                except OSError as e:
+                    LOG.exception(e)
 
     def _write_vrf_file(self, vrf_id, mapping_dict):
         return self._write_file(vrf_id, mapping_dict, self.vrf_mapping_file)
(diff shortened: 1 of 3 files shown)


opflexagent/gbp_ovs_agent.py
```diff
@@ -35,7 +35,6 @@
 from opflexagent import as_metadata_manager
 from opflexagent import constants as ofcst
 from opflexagent import rpc
-from opflexagent import snat_iptables_manager
```
amittbose: don't remove this line. SNAT stuff seems a little broken on 'master', it is missing this line: https://github.com/noironetworks/python-opflex-agent/blob/stable/juno/opflexagent/gbp_ovs_agent.py#L231

opflexagent/gbp_ovs_agent.py
```diff
@@ -274,15 +273,60 @@ def port_bound(self, port, net_uuid,
...
                     (physical_network in self.opflex_networks)):
                 # Port has to be untagged due to a opflex agent requirement
                 self.int_br.clear_db_attribute("Port", port.port_name, "tag")
-                self.mapping_to_file(port, net_uuid, mapping, fixed_ips,
+                # Multiple files will be created based on how many MAC
+                # addresses are owned by the specific port.
+                mapping_copy = copy.deepcopy(mapping)
+                mac_active_aap = {}
+                mapping_copy['allowed_address_pairs'] = []
+                mapping_copy['floating_ip'] = []
```
amittbose: Looks like mapping_copy['floating_ip'] will only contain the FIPs mapped to IPs specified in allowed-address-pairs. What about the FIPs that map to the regular fixed-IPs of the endpoint? Or did I miss something?

opflexagent/gbp_ovs_agent.py
```diff
@@ -274,15 +273,60 @@ def port_bound(self, port, net_uuid,
...
+                fip_by_fixed = {}
+                # Prepare FIPs by fixed_ip
+                for fip in mapping.get('floating_ip', []):
+                    fip_by_fixed.setdefault(
+                        fip['fixed_ip_address'], []).append(fip)
+                for aap in mapping.get('allowed_address_pairs', []):
+                    if aap.get('active'):
+                        if not aap.get('mac_address'):
```
amittbose: should also check the case where mac is present, but it is the same as the EP's mac

opflexagent/gbp_ovs_agent.py
```diff
@@ -274,15 +273,60 @@ def port_bound(self, port, net_uuid,
...
+                    mapping_copy['mac_address'] = mac
+                    # Extend the FIP list based on the allowed IPs
+                    mapping_copy['floating_ip'] = []
+                    mapping_copy['ip-address-mapping'] = []
+                    for aap in aaps:
+                        mapping_copy['floating_ip'].extend(fip_by_fixed.get(
+                            aap['ip_address'], []))
+                    # For this mac, set all the allowed address pairs.
+                    mapping_copy['allowed_address_pairs'] = aaps
+                    self.mapping_to_file(port, net_uuid, mapping_copy, None,
```
amittbose: Instead of None, pass [] for 'fixed_ips' parameter

opflexagent/gbp_ovs_agent.py
```diff
@@ -312,18 +356,24 @@ def mapping_to_file(self, port, net_uuid, mapping, fixed_ips,
...
             "interface-name": port.port_name,
-            "mac": mapping.get('mac_address') or port.vif_mac,
+            "mac": mac,
             "promiscuous-mode": mapping.get('promiscuous_mode') or False,
             "uuid": port.vif_id,
             'neutron-network': net_uuid}
 
-        ips = [x['ip_address'] for x in fixed_ips]
+        # Allocated IPs should be filtered by ownership.
+        if 'owned_addresses' in mapping:
```
amittbose: I  think owned_addresses is just the subset of allowed-addresses that is currently owned ("active"); regular fixed_ips are not included in it.

requirements.txt
```diff
@@ -7,4 +7,4 @@ oslo.utils<1.5.0,>=1.4.0 # Apache-2.0
 oslo.i18n<1.6.0,>=1.5.0 # Apache-2.0
 oslo.log<1.1.0,>=1.0.0 # Apache-2.0
 oslo.config<1.10.0,>=1.9.3 # Apache-2.0
-pyinotify
+#pyinotify
```
amittbose: This is needed to run UTs

mandeepdhami: Only on OSX ...

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Oh this is related to a different changeset. Apparently in Kilo Neutron doesn't pass down the root helper anymore. I'll fix this in master asap

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: good catch! Will do

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: I had the discussion with Mandeep yesterday, and we are actually also considering ownership of regular fixed ips. In which case we do not set up the IP in the EP file but still have DHCP.

requirements.txt
ivar-lazzaro: Yeah I forgot it commented out, will fix

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: ok

opflexagent/gbp_ovs_agent.py
amittbose: I think it is better to keep regular fixed-IPs separate from owned IPs as far as the return value of the RPC goes. The agent can process this in whichever way it likes.

I'm not sure how removing the IP-list from EP file is going to work. Also, what about no DHCP case? I think we should keep that change separate from the one at hand.

opflexagent/gbp_ovs_agent.py
amittbose: I thought I already fixed the root helper issue on Kilo, see https://github.com/noironetworks/python-opflex-agent/commit/9c5a7d9a2842b450b09bc15d3883ac7ea8d4afb5

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Fixed IP list is not currently filtered on the controller side. The agent can decide what to do with it.
I agree that we should know how the opflex agent will work if a DHCP mapping is not present in the EP list, and testing from this feature branch seems to be the quicker way to verify this.

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: It was still breaking the UTs, and I misunderstood the reason why this was happening. I've posted a patch in master that should fix it

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Done

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,70 @@ def port_bound(self, port, net_uuid,
...
+                        fip_by_fixed.get(fixed['ip_address'], []))
+
+                # For the main MAC EP, set al the AAP with no mac address or
+                # MAC address equal to the original MAC.
+                for aap in mapping.get('allowed_address_pairs', []):
+                    if aap.get('active'):
+                        if not aap.get('mac_address') or aap.get(
+                                'mac_address') == original_mac:
+                            # Should go with the MAIN mac address EP file
+                            mapping_copy['allowed_address_pairs'].append(aap)
```
amittbose: This doesn't look right. "allowed_address_pairs" maps to the "virtual-ip" entry in EP file. On the ovs-agent side, virtual-IP is interpreted as the IPs for which gratuitous ARPs needs to be trapped. So, allowed_address_pairs for the main mapping file should have all the allowed-addresses, including inactive ones.

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,70 @@ def port_bound(self, port, net_uuid,
...
+                            # Also set the right floating IPs
+                            mapping_copy['floating_ip'].extend(
+                                fip_by_fixed.get(aap['ip_address'], []))
```
amittbose: Also, add aap['ip_address'] to 'fixed_ips', or is that supposed to be done by the controller?

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,70 @@ def port_bound(self, port, net_uuid,
...
+                        else:
+                            # Store for future processing
+                            mac_active_aap.setdefault(
+                                aap['mac_address'], []).append(aap)
+                # Create mapping file for base MAC address
+                self.mapping_to_file(port, net_uuid, mapping_copy, fixed_ips,
                                      device_owner)
+                # Reset for AAP EP files
+                mapping_copy['allowed_address_pairs'] = []
+                mapping_copy['fixed_ips'] = []
```
amittbose: Probably safer to remove the key fixed_ips, extra_ips etc instead of setting them to empty list.

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,70 @@ def port_bound(self, port, net_uuid,
...
+                mapping_copy['subnets'] = []
+                mapping_copy['enable_dhcp_optimization'] = False
+                mapping_copy['enable_metadata_optimization'] = False
+                # Map to file based on the active AAP with a MAC address
+                for mac, aaps in mac_active_aap.iteritems():
+                    # Replace the MAC address with the new one
+                    mapping_copy['mac_address'] = mac
+                    # Extend the FIP list based on the allowed IPs
+                    mapping_copy['floating_ip'] = []
+                    mapping_copy['ip-address-mapping'] = []
```
amittbose: 'ip-address-mapping' is the wrong key, it should be 'ip_mapping'. Anyways, I thought about this a little, and I think it is fine to copy ip_mapping information from the main EP file to the additional EP file. That will allow SNAT for packets sent with the allowed-address.

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,70 @@ def port_bound(self, port, net_uuid,
...
+                    for aap in aaps:
+                        mapping_copy['floating_ip'].extend(fip_by_fixed.get(
+                            aap['ip_address'], []))
+                    # For this mac, set all the allowed address pairs.
+                    mapping_copy['allowed_address_pairs'] = aaps
```
amittbose: mapping_copy['allowed_address_pairs'] should be empty. Instead, the IP-addresses from 'aaps' needs to be copied to be provided as 'fixed_ips'

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: I see, that makes sense. this be true also for the other EP files, not only for the main one, is that correct?
Also, by "should have all the addresses" you mean those with no-mac or with mac==main_mac, right?

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Not sure what you mean, this gets the right floating IP based on the AAPs fixed ips.

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Why is that? The agent should be able to handle empty lists (also, avoids some unwanted KeyErrors).

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Would copying the info *as is* be a problem? Or should that be filtered as well?

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: What is the difference? Why should we promote AAPs to Fixed IPs for the additional files? No DHCP will be running for them anyways. Is this somehow a opflex-agent requirement?

opflexagent/gbp_ovs_agent.py
amittbose: Actually, I was more concerned about mapping_to_file() not behaving correctly. For example, the EP file produced by setting key fixed_ips to [] is very different from that produced by not having the key at all.

opflexagent/gbp_ovs_agent.py
amittbose: Unless these IPs appear in the primary IP list in the EP file, port-security rules in the OVS pipeline will discard the packets originating from the allowed mac+IP. Similarly, packets destined to the allowed-address won't be delivered.
Adding the virtual-ip section to this additional EP-file is really not needed, and it might actually confuse the OVS pipeline.

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,68 @@ def port_bound(self, port, net_uuid,
...
+                    mapping_copy['floating_ip'].extend(
+                        fip_by_fixed.get(fixed['ip_address'], []))
+
+                # For the main MAC EP, set al the AAP with no mac address or
+                # MAC address equal to the original MAC.
+                for aap in mapping.get('allowed_address_pairs', []):
+                    if not aap.get('mac_address') or aap.get(
+                            'mac_address') == original_mac:
+                        # Should go with the MAIN mac address EP file
+                        mapping_copy['allowed_address_pairs'].append(aap)
```
amittbose: For the main EP file, 'allowed_address_pairs' should NOT be filtered at all (because it maps to virtual-ip)

opflexagent/gbp_ovs_agent.py
```diff
@@ -277,15 +277,68 @@ def port_bound(self, port, net_uuid,
...
+                        # Also set the right floating IPs
+                        mapping_copy['floating_ip'].extend(
+                            fip_by_fixed.get(aap['ip_address'], []))
+                    else:
+                        # Store for future processing
+                        mac_aap_map.setdefault(
```
amittbose: Additional EP file needs to be created only for 'active' allowed-addresses.

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: My understanding was that we need a file for each MAC address. The "filtered" allowed address pairs here will just end up in a different File's "virtual-ip" section. How is this not equivalent of having one file with all the allowed address pairs clumped in the virtual-ip section?

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: I see. To avoid many back and forth, what is the list of keys that should be missing instead of being empty?

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: In the current implementation, allowed_address_pairs only end up as virtual-ip and not in the main IP file [0].
Am I missing something? Was this just broken from the start?

[0] https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/gbp_ovs_agent.py#L347-L351

opflexagent/gbp_ovs_agent.py
amittbose: fixed_ips and extra_ips I think should be good enough

opflexagent/gbp_ovs_agent.py
amittbose: Not sure I understand your question. AAP just map to virtual-IP, and that is why they need to be promoted to fixed IP when a particular AAP is 'active'. If the AAP has a different MAC, then the IP-address appears in the fixed IPs for a different file; else it appears in the fixed IPs for the main file.

As for virtual-IP section in the additional EP file, my concern was that it may cause problems with managing lifecycle of flows added for the virtual-IPs. This may or may not be an issue, so I simply wanted to caution you.

opflexagent/gbp_ovs_agent.py
amittbose: Are you creating an EP file for an inactive AAP with a different MAC? If yes, then it'll probably be fine. This is kind of confusing though because the additional EP file is really for an AAP (=virtual-ip), and having another virtual-IP section in it is just odd.
Note that if you are creating this file for inactive AAPs, then the promotion of AAP to fixed-IP needs to happen only for active AAPs.

ivar-lazzaro: Yes I am. Please refer to the UTs added below (especially the multi file one, in which I pretty much cover all this).
Consider that multiple AAP may and up in the same file (AAP with same mac address) so you basically store them by MAC (one or multiple).

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Thanks! Will do.

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: Sorry I wasn't clear with my question:

Given that so far allowed_address_pairs only ended up filling the virtual-ip section... How did the port filtering work? How could a VM even send traffic from a AAP if the said address was never set in the "ip" section? Was it broken or is the opflex agent actually doing port security also considering the virtual-ip section?

As for the additional files, I believe that one or multiple files with their own virtual-ip sections should work seamlessly. The notion of "main" EP file should only exist in Openstack (artifact of a Neutron concept). Also consider that the virtual-ip sections of various EP files will never ovarlap, since they are always clustered by MAC address.

amittbose: The AAP/VIP support was very much a work in progress. In your changes are very much part of the changes needed to support AAP. Sorry, I didn't realize that you were not aware of that :).

As far as the ovs agent goes, the only way it will allow an EP to send/receive traffic is if the IP is listed in the fixed-IP. The job of the opflex-agent is to promote an AAP to fixed-IP when the AAP becomes active; this will enable the EP to send/receive traffic using the AAP.
On the other hand, the virtual-IP section is meant for detection of failovers - it is simply used to trap grat ARPs and notify that the ARP-ed address has become active. This update goes up to the controller, and the controller notifies the opflex-agent to re-generate the EP file. And a new file is expected to be created where the ARP-ed IP (i.e. an allowed IP) is promoted to be in the fixed-IP.

ivar-lazzaro: Thanks! I had no idea the AAP addresses had to be on the IP section for filtering. I only based my implementation on the existing code and that was misleading on this point. Will provide the patch asap.

PR #16, opened 2015-11-04
add extra details per mac address when needed

ivar-lazzaro:
Use Case: Head of the chain is active and part of a cluster, all
the backend IP information must go to the right MAC address
(server will do the proper choice).


ivar-lazzaro:
Let's start the reviews, but we may want to wait a little before merging

PR #26, opened 2016-02-15
Added support for DVS/OVS OpFlex Agent

tbachman:
This adds a derived agent that provides a new parameter
to the plugin, hypervisor_type, which can be used by the
mechanism driver to indicate the appropriate VIF type to
nova when peroforming port binding. It also adds a specialized
Endpoint manager and Firewall driver, in order to support
multiple firewall implementations, and prevent creation of
EP files for ports owned by compute:nova on hosts that are
running nova-compute for vCenter.

Signed-off-by: Thomas Bachman <tbachman@yahoo.com>

### DIFF
--- a/opflexagent/config.py
+++ b/opflexagent/config.py
@@ -30,6 +30,16 @@ gbp_opts = [
                 default=['*'],
                 help=_("List of the physical networks managed by this agent. "
                        "Use * for binding any opflex network to this agent")),
+    cfg.ListOpt('firewall_map',
+                default=[('dvs_port_key',
+                    'opflexagent.utils.fw_drivers.dvs.DVSFirewallDriver'), ],
+                help=_("List of tuples containing firewall driver and the "
+                       "key that selects it")),
+    cfg.StrOpt('hypervisor_type',
+               default='',
+               help=_("Hypervisor of the host running this agent. Only needed "
+                      "on hosts running nova-compute. Set to 'VMware vCenter' "
+                      "for hosts running nova-compute for vCenter")),
     cfg.ListOpt('internal_floating_ip_pool',
                default=['169.254.0.0/16'],
                help=_("IP pool used for intermediate floating-IPs with SNAT")),
--- a/opflexagent/firewall_wrapper.py
+++ b/opflexagent/firewall_wrapper.py
@@ -1,0 +1,181 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+import contextlib
+from neutron.agent import firewall as fw
+from oslo_log import log as logging
+from oslo_utils import importutils
+
+
+LOG = logging.getLogger(__name__)
+
+
+class FirewallDriverWrapper(fw.FirewallDriver):
+    """Wrapper for Firewall Driver implemntations
+
+    This wrapper class keeps a map of strategies to
+    use for implementing a firewall driver. The strategy
+    selected is determined by a key, which is passed in
+    a map as part of the init_wrapper function. The strategy
+    must know how to select the appropriate firewall driver
+    based solely on the key associated with that driver. The
+    key should be obtained from the port object, or if no port
+    object is supplied, it should return all of them
+    """
+
+    def __init__(self):
+        self.strategy_map = {}
+        self.strategy_fn = None
+        self.default_strategy = None
+
+    def strategy(self, port):
+        """Return appropiate strategy object
+
+        This returns the appropriate implementation of
+        a FirewallDriver, based on the strategy function
+        and strategy map provided. If no strategy object
+        is found, then the default strategy object is
+        returned. This is only valid for port-based
+        API calls, which can select the right strategy
+        based on port properties.
+        """
+        strategy_obj = self.default_strategy
+
+        if self.strategy_map and self.strategy_fn:
+            key = self.strategy_fn(port, self.strategy_map)
+            if key:
+                strategy_obj = self.strategy_map[key]
+        return strategy_obj
+
+    def set_strategy_and_map(self, strategy_map, strategy_fn, default):
+        """Set Firewall Driver strategy and map
+
+           This sets the strategy to call for selecting
+           which firewall driver to use. It also provides
+           the map the strategy uses for selecting it.
+        """
+        if not strategy_map and default is None:
+            return
+
+        # set default strategy if a map wasn't provided
+        if not strategy_map:
+            self.default_strategy = importutils.import_object(default)
+
+        self.strategy_fn = strategy_fn
+
+        # We need to go through and load each of the classes
+        # in the firewall map and provide that in our map
+        for element in strategy_map:
+            key, value = element
+            self.strategy_map[key] = importutils.import_object(value)
+
+    def prepare_port_filter(self, port):
+        self.strategy(port).prepare_port_filter(port)
+
+    def apply_port_filter(self, port):
+        self.strategy(port).apply_port_filter(port)
+
+    def update_port_filter(self, port):
+        self.strategy(port).update_port_filter(port)
+
+    def remove_port_filter(self, port):
+        self.strategy(port).remove_port_filter(port)
+
+    # The following methods apply across all strategies
+    def filter_defer_apply_on(self):
+        """Turn on deferred apply
+
+           This method is not port-based, and therefore must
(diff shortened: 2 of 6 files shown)


opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+import contextlib
+from neutron.agent import firewall as fw
+from oslo_log import log as logging
+from oslo_utils import importutils
+
+
+LOG = logging.getLogger(__name__)
+
+
+class FirewallDriverWrapper(fw.FirewallDriver):
```
ivar-lazzaro: Should this be called Factory instead of Wrapper?

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+    """Wrapper for Firewall Driver implemntations
+
+    This wrapper class keeps a map of strategies to
+    use for implementing a firewall driver. The strategy
+    selected is determined by a key, which is passed in
+    a map as part of the init_wrapper function. The strategy
```
ivar-lazzaro: couldn't find the init_wrapper method, am I missing something?

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+            if key:
+                strategy_obj = self.strategy_map[key]
+        return strategy_obj
+
+    def set_strategy_and_map(self, strategy_map, strategy_fn, default):
+        """Set Firewall Driver strategy and map
+
+           This sets the strategy to call for selecting
+           which firewall driver to use. It also provides
+           the map the strategy uses for selecting it.
```
ivar-lazzaro: s/uses/used

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+    key should be obtained from the port object, or if no port
+    object is supplied, it should return all of them
+    """
+
+    def __init__(self):
+        self.strategy_map = {}
+        self.strategy_fn = None
+        self.default_strategy = None
+
+    def strategy(self, port):
```
ivar-lazzaro: is this meant to be a public method? Looks like something you would just use internally

opflexagent/firewall_wrapper.py
ivar-lazzaro: After a better look, this is not really a factory, wrapper it's more appropriated.

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+           This sets the strategy to call for selecting
+           which firewall driver to use. It also provides
+           the map the strategy uses for selecting it.
+        """
+        if not strategy_map and default is None:
+            return
+
+        # set default strategy if a map wasn't provided
+        if not strategy_map:
+            self.default_strategy = importutils.import_object(default)
```
ivar-lazzaro: why only if a map is not provided? Isn't default_strategy also used in case of an invalid/unknown strategy?

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,181 @@
...
+
+        self.strategy_fn = strategy_fn
+
+        # We need to go through and load each of the classes
+        # in the firewall map and provide that in our map
+        for element in strategy_map:
+            key, value = element
+            self.strategy_map[key] = importutils.import_object(value)
```
ivar-lazzaro: I see that there are two kind of methods following this point: "apply with strategy" and "apply all strategies". My experience in this case (especially with Openstack) is that for each new resource/API added to a driver we end up duplicating a lot of this code... Which is error prone and all that bad stuff.

My suggestion here would be to either have two private functions and make one liner overrides (eg: _apply_on_strategy/_apply_all_strategies where you pass down the method name to be called on the driver), or to use some __getattr__ python black magic to automatically manage the public methods (this is my favorite, but very debatable). Obviously there might be other ways.

opflexagent/gbp_ovs_agent.py
```diff
@@ -112,6 +123,20 @@ def __init__(self, root_helper=None, **kwargs):
         # TODO(ivar): make this component pluggable.
         self.ep_manager = ep_manager.EndpointFileManager().initialize(
             self.host, self.bridge_manager, kwargs)
+        self.init_firewall()
+
+    def _invoke_ep_mgr(self, port):
```
ivar-lazzaro: what about loading a void EP manager when VCenter is used? this would reduce by a decent amount the code change in this class, which is the most overloaded one, by adding an EP manager that simply inherits from the base and does basically nothing.

opflexagent/gbp_ovs_agent.py
```diff
@@ -531,6 +561,40 @@ def set_rpc_timeout(self, timeout):
...
+           default strategy will be used (defined by the
+           SECURITYGROUP config file)
+        """
+        firewall_driver = cfg.CONF.SECURITYGROUP.firewall_driver
+        LOG.debug("Init firewall settings (driver=%s)", firewall_driver)
+        if not sg_rpc._is_valid_driver_combination():
+            LOG.warn(_LW("Driver configuration doesn't match "
+                         "with enable_security_group"))
+        self.firewall_map = cfg.CONF.OPFLEX.firewall_map
+        self.firewall = importutils.import_object(
```
ivar-lazzaro: why not instantiating the wrapper directly?

opflexagent/config.py
```diff
@@ -30,6 +30,16 @@
                 default=['*'],
                 help=_("List of the physical networks managed by this agent. "
                        "Use * for binding any opflex network to this agent")),
+    cfg.ListOpt('firewall_map',
```
ivar-lazzaro: this firewall map looks a lot like a list of driver entry points. I would suggest we follow that kind of implementation that is widely used in openstack... Looking at how ML2 works with mechanism drivers is a good start! We can have a webex around this

opflexagent/firewall_wrapper.py
tbachman: Nope -- I am :-). I'll change this to "set_strategy_and_map" (but only if we keep this class -- see additional comments).

opflexagent/firewall_wrapper.py
tbachman: will fix

opflexagent/firewall_wrapper.py
tbachman: Good point -- I'll prepend with an underscore.

opflexagent/gbp_ovs_agent.py
tbachman: That was my original  implementation. However, it requires a user-configurable parameter. I received feedback earlier that we should try to avoid new configuration parameters if possible, and keep a single implementation that supports multiple things. I'm certainly open to revisiting this, but wanted to provide context for the current implementation.

opflexagent/firewall_wrapper.py
tbachman: Okay - I'll take a stab at cleaning that up (I'd thought about using getattr to simplify testing as well)

opflexagent/config.py
tbachman: That's a good point (no sense in re-inventing the wheel ;-)!  I'll see what code can be leveraged there. The main thing I need is the ability to select an implementation based on a strategy, where the strategy method returns a key that selects the strategy. I think they use some of the stevedore libs there, so I'll look if I can simplify the code some by leveraging that.

opflexagent/firewall_wrapper.py
tbachman: The difference with the default_strategy is that you don't need a key, and therefore can bypass any strategy. I guess this could be changed to not invoke the strategy if the map has only a single entry. Does that seem better than using a default?

opflexagent/gbp_ovs_agent.py
tbachman: Good question. I have no idea :-). I'll change it to use it directly.

tbachman:
I also realized that the security groups wrapper isn't connected (needs to be tied to the sg_agent member, and not the current firewall member). Will close this and redo with a new patch.

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: hypervisor_type is already a user configurable parameter, we could use it to chose which manager to use.

tbachman: Just thinking about your TODO in the code -- do we anticipate other types of EP managers? If so, then I guess we are already in the world of subclassing, so it makes more sense to do it directly here (although then you can have misconfigurations -- DVS w/wrong EP manager... but there's the other pitfall of tying it to hypervisor type, but then not allowing someone to override the EP manager if they want).

ivar-lazzaro:
After rebasing/changing your feature branch on master, you can use "push -f" in your private branch (!!!) in order to keep the commit history clean and incrementally update the same pull request... By deleting and recreating the branch/pr we lose all the review comments.

tbachman:
Got it -- wasn't sure what the convention/protocol was here, so I'll do that next time.  Thanks!

tbachman:
FYI - I rebased the current patch (for your review):
https://github.com/noironetworks/python-opflex-agent/pull/28

PR #28, opened 2016-02-17
Added support for DVS/OVS OpFlex Agent

tbachman:
This adds a derived agent that provides a new parameter
to the plugin, hypervisor_type, which can be used by the
mechanism driver to indicate the appropriate VIF type to
nova when peroforming port binding. It also adds a specialized
Endpoint manager and Firewall driver, in order to support
multiple firewall implementations, and prevent creation of
EP files for ports owned by compute:nova on hosts that are
running nova-compute for vCenter.

Signed-off-by: Thomas Bachman <tbachman@yahoo.com>

### DIFF
--- a/opflexagent/config.py
+++ b/opflexagent/config.py
@@ -30,6 +30,16 @@ gbp_opts = [
                 default=['*'],
                 help=_("List of the physical networks managed by this agent. "
                        "Use * for binding any opflex network to this agent")),
+    cfg.ListOpt('firewall_map',
+                default=[('dvs_port_key',
+                    'opflexagent.utils.fw_drivers.dvs.DVSFirewallDriver'), ],
+                help=_("List of tuples containing firewall driver and the "
+                       "key that selects it")),
+    cfg.StrOpt('hypervisor_type',
+               default='KVM',
+               help=_("Hypervisor of the host running this agent. Only needed "
+                      "on hosts running nova-compute. Set to 'VMware vCenter' "
+                      "for hosts running nova-compute for vCenter")),
     cfg.ListOpt('internal_floating_ip_pool',
                default=['169.254.0.0/16'],
                help=_("IP pool used for intermediate floating-IPs with SNAT")),
--- a/opflexagent/firewall_wrapper.py
+++ b/opflexagent/firewall_wrapper.py
@@ -1,0 +1,115 @@
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from oslo_log import log as logging
+from oslo_utils import importutils
+
+
+LOG = logging.getLogger(__name__)
+
+
+class FirewallDriverWrapper(object):
+    """Wrapper for Firewall Driver implemntations
+
+    This wrapper class keeps a map of drivers to use for
+    implementing a firewall driver. The firewall driver methods
+    are divided into two groups:
+      1) methods where only one driver implementation should be called
+      2) methods where all driver implementations should be called
+
+    The first group is called when there is only a single argument:
+    the neutron port. The second group is called in all other cases.
+    The first group uses a key to determine which driver gets invoked.
+    The key is obtained by a user-defined strategy function, which gets
+    passed the port object. The strategy function is meant to perform
+    some operation (e.g. search) on the port to determine and return
+    the key to use for looking up the appropriate firewall driver.
+
+    There is also a defaulit firewall driver. This is used when there
+    is only a single firewall driver is used, and therefore a key is
+    not needed.
+    """
+
+    def __init__(self):
+        self._strategy_map = {}
+        self._strategy_fn = None
+        self._default_strategy = None
+        self._port_methods = ['prepare_port_filter',
+                              'apply_port_filter',
+                              'update_port_filter',
+                              'remove_port_filter']
+        self._all_strategy_methods = ['filter_defer_apply_on',
+                                      'filter_defer_apply_off',
+                                      'ports',
+                                      'defer_apply',
+                                      'update_security_group_members',
+                                      'update_security_group_rules']
+
+    def _do_port_strategy(self, name, port, *p_args, **p_kwargs):
+        """Call a single firewall driver method"""
+        strategy_obj = self._default_strategy
+        if self._strategy_map and self._strategy_fn:
+            key = self._strategy_fn(port, self._strategy_map)
+            if key:
+                strategy_obj = self._strategy_map[key]
+        return getattr(strategy_obj, name)(port)
+
+    def _do_all_strategy(self, name, *s_args, **s_kwargs):
+        """Call all the firewall driver methods"""
+        if self._default_strategy:
+            return getattr(self._default_strategy, name)(*s_args, **s_kwargs)
+        else:
+            ret_list = []
+            for driver in self._strategy_map.values():
+                ret_list.append(getattr(driver, name)(*s_args, **s_kwargs))
+            return ret_list
+
+    def __getattr__(self, name):
+        def _do_strategy(*args, **kwargs):
+            """Invoke the appropiate strategy method
+
+            This checks whether the method is in the port-based
+            methods or non-port-based methods, and invokes the
+            appropriate strategy. If it's not a valid method,
+            it raises AttributeError
+            """
+            if name in self._port_methods:
+                return self._do_port_strategy(name, *args, **kwargs)
+            elif name in self._all_strategy_methods:
+                return self._do_all_strategy(name, *args, **kwargs)
+            else:
+                raise AttributeError
+        return _do_strategy
+
+    def set_strategy_and_map(self, strategy_map, strategy_fn, default):
+        """Set Firewall Driver strategy and map
+
+           This sets the strategy to call when selecting
+           which firewall driver to use. It also provides
+           a list of tuples, which are turned into a map
+           of security group objects/implementations, indexed
+           by their keys.
+        """
+        if not strategy_map and default is None:
+            return
+
+        # set default strategy if a map wasn't provided
+        if not strategy_map:
+            self._default_strategy = importutils.import_object(default)
+
+        self._strategy_fn = strategy_fn
+
+        # We need to go through and load each of the classes
+        # in the firewall map and provide that in our map
+        for element in strategy_map:
+            key, value = element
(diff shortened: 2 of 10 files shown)


tbachman:
I need to redo the EP unit tests - not the right way to do it. Will resubmit.

tbachman:
After thinking about this some more, I didn't come up with a good way to reuse code for the UTs (i.e. the logic of "just do the opposite" doesn't always apply -- e.g. when we actually want to assert something isn't called, then we don't want to reverse it). Unless others have a clever way of generalizing this, I'm going to leave this as-is (i.e. go ahead and review).

opflexagent/config.py
```diff
@@ -30,6 +30,16 @@
                 default=['*'],
                 help=_("List of the physical networks managed by this agent. "
                        "Use * for binding any opflex network to this agent")),
+    cfg.ListOpt('firewall_map',
+                default=[('dvs_port_key',
+                    'opflexagent.utils.fw_drivers.dvs.DVSFirewallDriver'), ],
+                help=_("List of tuples containing firewall driver and the "
+                       "key that selects it")),
+    cfg.StrOpt('hypervisor_type',
+               default='',
```
rkukura: Should there be some default value to indicate non-vCenter hypervisors, such as "KVM" or "normal"?

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+    implementing a firewall driver. The firewall driver methods
+    are divided into two groups:
+      1) methods where only one driver implementation should be called
+      2) methods where all driver implementations should be called
+
+    The first group is called when there is only a single argument:
+    the neutron port. The second group is called in all other cases.
+    The first group uses a key to determine which driver gets invoked.
+    The key is obtained by a user-defined strategy function, which gets
+    passed the port object. The strategy function is meant ot perform
```
rkukura: "meant ot perform" -> "meant to perform"

opflexagent/config.py
```diff
@@ -30,6 +30,16 @@
                 default=['*'],
                 help=_("List of the physical networks managed by this agent. "
                        "Use * for binding any opflex network to this agent")),
+    cfg.ListOpt('firewall_map',
+                default=[('dvs_port_key',
```
rkukura: Is this supposed to be "dvs_port_group" rather than "dvs_port_key"?

opflexagent/config.py
tbachman: I think we'll have the following cases:
1) not present (i.e. no "hypervisor_type" property in config). This is for interoperability with old agents, which don't have this code. The behavior here defaults to KVM.
2) explicitly populated with "VMware vCenter". This is the ESX use case (i.e. the reason for this patch)
3) explicitly populated with "KVM". This should be the default, and I'll amend this pull-request to reflect that. Interoperability with old ML2 drivers works as they will ignore this, but it could possibly be used in the future by new versions of the driver, if needed.
4) Other possible values (e.g. bare metal). This would allow us to extend it's usage. We could define "None" now, but maybe it's best to wait until we implement that use case to decide (note: whatever it is, it must be explicit, as the above already defines the implicit behavior when this property is absent).

opflexagent/firewall_wrapper.py
tbachman: will fix

opflexagent/config.py
tbachman: This is meant to be the port  key -- this value is the key used to select the DVS firewall driver, and not the port group (port groups will be dynamic, whereas the firewall drivers will be configured).

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,102 @@
...
+            ret_list = []
+            for driver in self._strategy_map.values():
+                ret_list.append(getattr(driver, name)(*s_args, **s_kwargs))
+            return ret_list
+
+    def __getattr__(self, name):
+        def _do_strategy(*args, **kwargs):
+            """Invoke the appropiate strategy method
+
+            This checks the number of arguments, and if 1,
```
ivar-lazzaro: Is this check enough? would two sets of explicitly mapped methods be a safer approach?
something like:

if name in self.port_methods:
    self._do_port_strategy(name, *args, **kwargs)
elif name in self.all_strategy_methods:
    self._do_all_strategy(name, *args, **kwargs)
else:
    raise AttributeError

opflexagent/gbp_ovs_agent.py
```diff
@@ -72,6 +79,7 @@ def __init__(self, root_helper=None, **kwargs):
         self.root_helper = root_helper
         self.notify_worker = opflex_notify.worker()
         self.host = cfg.CONF.host
+        self.hypervisor_type = cfg.CONF.OPFLEX.hypervisor_type
```
ivar-lazzaro: Add this to create_agent_config_map

opflexagent/gbp_ovs_agent.py
```diff
@@ -110,8 +121,12 @@ def __init__(self, root_helper=None, **kwargs):
         self.quitting_rpc_timeout = kwargs['quitting_rpc_timeout']
         # Initialize the Endpoint Manager.
         # TODO(ivar): make this component pluggable.
-        self.ep_manager = ep_manager.EndpointFileManager().initialize(
-            self.host, self.bridge_manager, kwargs)
+        if (self.hypervisor_type and
```
ivar-lazzaro: when self.hypervisor_type is not None, it's also not equal to ofcst.HYPERVISOR_VCENTER so you can remove the first part of the check

opflexagent/gbp_ovs_agent.py
```diff
@@ -546,6 +561,43 @@ def set_rpc_timeout(self, timeout):
             rpc_api.client.timeout = timeout
 
 
+class OpflexSecurityGroupAgentRpc(sg_rpc.SecurityGroupAgentRpc):
```
ivar-lazzaro: would this be better placed in the rpc.py file?

opflexagent/firewall_wrapper.py
tbachman: Definitely. Will fix (and thanks).

opflexagent/gbp_ovs_agent.py
tbachman: Oops. Thanks, and good catch. Will fix.

opflexagent/test/test_endpoint_file_manager.py
```diff
@@ -563,3 +574,200 @@ def test_registered_endpoints(self):
             manager.declare_endpoint(port, mapping)
             self.assertEqual(set(['uuid2', port.vif_id]),
                              manager.get_registered_endpoints())
+
+
+class TestEndpointFileManagerDvs(TestEndpointFileManager):
```
ivar-lazzaro: should this inherit from TestEndpointFileManagerBase instead?

opflexagent/gbp_ovs_agent.py
tbachman: Will fix.

opflexagent/gbp_ovs_agent.py
tbachman: Seems like a good idea. Will fix.

opflexagent/test/test_endpoint_file_manager.py
tbachman: My goal was to have all of the same tests as the non-DVS use case, except have them be tested with non-compute EPs (the DVS version should behave the same for non-compute EPs). I was using this approach to try and minimize code duplication, and still have these tests. I am open to changing this if it makes more sense to have it inherit from TestEndpointFileManagerBase.

tbachman:
FYI -- I updated the pull request based on the feedback.

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,115 @@
...
+            invokes the key-based strategy. If not 1, it calls
+            all the drivers.
+            """
+            if name in self._port_methods:
+                return self._do_port_strategy(name, args[0])
```
ivar-lazzaro: you don't know if there's only one argument in this case, should still pass *args and **kwargs

opflexagent/firewall_wrapper.py
ivar-lazzaro: This docstring is outdated

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    def __init__(self):
+        self._strategy_map = {}
+        self._strategy_fn = None
+        self._default_strategy = None
+        self._port_methods = ['prepare_port_filter',
+                              'apply_port_filter',
+                              'update_port_filter',
+                              'remove_port_filter',
```
ivar-lazzaro: remove_port_filter is duplicated in this list.

opflexagent/firewall_wrapper.py
```diff
@@ -0,0 +1,115 @@
...
+                              'remove_port_filter']
+        self._all_strategy_methods = ['filter_defer_apply_on',
+                                      'filter_defer_apply_off',
+                                      'ports',
+                                      'defer_apply',
+                                      'update_security_group_members',
+                                      'update_security_group_rules']
+
+    def _do_port_strategy(self, name, port):
```
ivar-lazzaro: signature here should be something like:

def _do_port_strategy(self, name, port, *args, **kwargs):

opflexagent/gbp_ovs_agent.py
ivar-lazzaro: modify this into:
`self.hypervisor_type = kwargs['hypervisor_type']`

opflexagent/gbp_ovs_agent.py
```diff
@@ -550,6 +564,7 @@ def create_agent_config_map(conf):
     agent_config = ovs.create_agent_config_map(conf)
     agent_config['epg_mapping_dir'] = conf.OPFLEX.epg_mapping_dir
     agent_config['opflex_networks'] = conf.OPFLEX.opflex_networks
+    agent_config['hypervisor_config'] = cfg.CONF.OPFLEX.hypervisor_type
```
ivar-lazzaro: why hypervisor_config and not hypervisor_type?

opflexagent/firewall_wrapper.py
tbachman: Will fix.

opflexagent/gbp_ovs_agent.py
tbachman: will change.

opflexagent/firewall_wrapper.py
tbachman: will fix.

opflexagent/firewall_wrapper.py
tbachman: will fix.

opflexagent/firewall_wrapper.py
tbachman: will change.

opflexagent/gbp_ovs_agent.py
tbachman: Because I'm apparently ADD :-P.  Will fix, and good catch.

tbachman:
Abandoning -- using DVS agent instead.

PR #34, opened 2016-06-30
Transiction refactored agent to asynchronous model

ivar-lazzaro:
- Introduce Port Manager. This module takes care of managing
the async comunication between Neutron and the Opflex Agent
- Switch to the request_* RPC calls.

To avoid increasing the chance of conflicts even further, some
of the methods currently on the main agent class were not moved
to the port manager. This will be done eventually.


ivar-lazzaro:
Cherry picking was not trivial due to many merge conflicts, so I'm creating a new pull request for this.

PR #54, opened 2017-05-09
Evaluate the endpoint request timeout in seconds instead of milliseconds

ivar-lazzaro:


amittbose:
Please fix the pep8 failure before merge.

PR #55, opened 2017-05-24
Fix IPv6 subnets

tbachman:
The v6subnets was being treated as a list. It's actually a dictionary
whose keys are the subnet UUIDs.

close noironetworks/support#540


amittbose:
How about adding a unit-test for the v6 stuff?

tbachman:
Yes -- good point. Will add.

tbachman:
Added a UT for this and pushed update.

PR #56, opened 2017-06-05
Route leak feature support

kentwu1:
add the route-leak-list to the rdconfig file.


kentwu1:
this is not needed anymore.

PR #59, opened 2017-08-18
Stable/ocata for master

tbachman:
This branch will be merged to master, so that master can
track the stable/ocata branch


amittbose:
Can you also update the version in setup.py to 6.0.0?

tbachman:
Good call -- will chage.

snaiksat:
Hi Thomas, I also seemed to have added this note in my commit:
The ``dhcp_domain`` DHCP agent configuration option was deprecated in
Liberty cycle in Neutron, and now is now no longer used.
The ``dns_domain`` option should be used instead.

The idea was to highlight this change. If it's not too much work, could you add it to your commit message?

tbachman:
Definitely -- and thanks!

tbachman:
Pushed update to address reviewer comments.

PR #62, opened 2017-09-07
Add apic host and metadata proxy agents

tbachman:
The apic host agent and metadata namespace proxy agents used
to live in the apic-ml2-driver repo. This meant that the apic-ml2-driver
package and its dependencies had to be installed on compute nodes, even
though most of the apic-ml2-driver wasn't needed. This patch adds the
agents to this repo, as the agents will be deployed wherever there's
a neutron-opflex-agent.


rkukura:
Don't the executable paths in setup.py need to start with "opflexagent."?

tbachman:
You're right - I discovered this, but forgot to update setup.py. Good catch!

PR #66, opened 2017-09-22
Security group support

kentwu1:
1. Create the br-fabric bridge during initialization.
2. Create/Delete the patch ports on br-int and br-fabric respectively
when a VM is being created/deleted.
3. adding the patch ports + SG info to the EP file.


kentwu1:
Let's review this but don't merge it yet until Ebro-MR1 is released.

kentwu1:
Need to push a new revision to use br-fabric instead in order to support the upgrade process better.

kentwu1:
pushed a new revision to use br-fabric approach instead. Note that we should just review this but don't merge it yet until Ebro-MR1 is released.

kentwu1:
pushed a revision to address the review comments.

kentwu1:
Ivar said he has approved this patch so merging it now.

PR #68, opened 2017-09-26
neutron_opflex_agent process on restart is changing the oflow protoco…

ratnakark:
neutron_opflex_agent process on restart is changing the oflow protocol  supported on br-int

replace impl_vsctl.BaseCommand which was not working with ovsdb set_protocols call


opflexagent/utils/bridge_managers/ovs_manager.py
```diff
@@ -51,7 +51,8 @@ def setup_integration_bridge(self):
         """Override parent setup integration bridge."""
         self.int_br.create()
         self.int_br.set_secure_mode()
-        self.int_br.reset_ofversion()
+        self.int_br.set_protocols(protocols='[]')
+        #self.int_br.reset_ofversion()
```
ivar-lazzaro: nit: do we need to keep this like commented?

PR #69, opened 2017-09-27
Support upstream dns_domain

ivar-lazzaro:


tbachman:
Does this need a backport?

PR #73, opened 2017-10-26
Fix metadata host routes

tbachman:
The agent needs to provide the metata host route when optimized
metadata is used. This is needed in cases where the neutron router
isn't connected to the subnet, and therefore ARPs for the default
gateway don't succeed, precluding any attempts to communicate with
the metadata server. In these cases, the DHCP server IP is used
as the next hop (i.e. the same solution used in the reference
implementation for Neutron with isolated networks).


tbachman:
This fix needs to be made in the plugin (the "inject_default_routes" flag in the L2P for GBP workflow isn't available at the agent, and therefore it can't do the correct thing in all cases). Closing this pull request, and will create a patch for upstream GBP.

PR #74, opened 2017-11-02
Allow different pod_id in LLDP

kentwu1:


kentwu1:
yes good catch, will do.

kentwu1:
Pushed a revision to address the review comments.

kentwu1:
yes your right. Thx.

kentwu1:
Pushed a revision to address the review comments.

PR #81, opened 2017-12-08
Add support for network updates

tbachman:
The network update RPC is added, along with another RPC that requests
notifications for all ports the agent owns on that network. The network
update notification is used as a trigger for the agent, to let it
know that it needs to query for the ports on the network, as some
non-port parameters may have changed (e.g. DNS, DHCP, etc.), which
get delivered as part of the gbp_details information.


amittbose:
As per my discussion with Thomas offline, it seems that doing the RPCs in this way does not reduce the messages exchanged, but in fact increases them. In the short term, I think the server side should just send a 'port_update' message for all affected ports. In future, we could consider enhancing this message to also provide a list of port IDs instead of a single port.

tbachman:
Closing, as this patch is no longer necessary.

PR #84, opened 2018-01-10
Pike sync

snaiksat:


opflexagent/test/test_gbp_ovs_agent.py
```diff
@@ -96,8 +96,6 @@ def start(self, interval=0):
             mock.patch('neutron.agent.common.ovs_lib.OVSBridge.'
                        'get_local_port_mac',
                        return_value='00:00:00:00:00:01'),
-            mock.patch('neutron.agent.linux.utils.get_interface_mac',
```
tbachman: This is no longer needed?

snaiksat: The function was earlier deprecated and was eventually removed from Neutron. The UTs pass, hence presumably no substitute mock is required.

PR #93, opened 2018-02-01
Only handle bound ports

tbachman:
The RPC to get devices for the agent doesn't check to see
if the port is bound to the agent/host. This can lead to the
creation of EP files for ports that the agent isn't responsible
for. A check is added to the device's host. If no host is specified,
or if the host matches the host for the agent, then the port is
processed normally. If there is a host, and it doesn't match the
host for the agent, then the port is removed.

### DIFF
--- a/opflexagent/gbp_ovs_agent.py
+++ b/opflexagent/gbp_ovs_agent.py
@@ -353,7 +353,11 @@ class GBPOpflexAgent(sg_rpc.SecurityGroupAgentRpcCallbackMixin,
             if gbp_details and 'port_id' not in gbp_details:
                 # The port is dead
                 details.pop('port_id', None)
-            if neutron_details and 'port_id' in neutron_details:
+            if (gbp_details and gbp_details.get('host') and
+                gbp_details['host'] != self.host):
+                    self.port_unbound(device)
+                    return False
+            elif neutron_details and 'port_id' in neutron_details:
                 LOG.info(_("Port %(device)s updated. Details: %(details)s"),
                          {'device': device, 'details': details})
                 # Inject GBP/Trunk details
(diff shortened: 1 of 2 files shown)


amittbose:
Did you verify that live migration of VMs works with this change?

tbachman:
Yeah -- verified that the instance can ping the subnet gateway and DHCP agent after migraiton.

opflexagent/gbp_ovs_agent.py
```diff
@@ -353,7 +353,10 @@ def treat_devices_added_or_updated(self, details):
             if gbp_details and 'port_id' not in gbp_details:
                 # The port is dead
                 details.pop('port_id', None)
-            if neutron_details and 'port_id' in neutron_details:
+            if (gbp_details and gbp_details.get('host') and
+                gbp_details['host'] != self.host):
+                    self.port_unbound(device)
```
amittbose: should 'return False' here?

tbachman: Good point -- when I implemented this, I followed what was done for this case:

https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/gbp_ovs_agent.py#L380-L383

In this case, the EP isn't defined and the EP file is removed. However, it still returns True, which causes the EP to be added to the list. What's not clear is if that's intentional, and if so, whether the above behavior should be the same. To me, it seems like both this case, and the one in my patch, should both return False. I'll make this change and test it out functionally, but am also curious if you see any problem with this.

tbachman: It looks like the original intent of that other code path was to not have anything added to the skipped devices list:

https://github.com/noironetworks/python-opflex-agent/commit/f23b3e27761b8eaf68848ca92a1268af52ba197a#diff-94e4cc6090ff079d0f2ff2bac60d2ad9R274

I'll add returns of false for both these cases in a new spin of the patch.

tbachman:
Pushed a new version of the patch. Functionally tested HA LBaaS and live-migration, and both worked.

PR #97, opened 2018-02-19
Add missing .ini file

tbachman:
The ml2_conf_cisco_apic.ini file was missed during its
port from the apic-ml2-driver package.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,5 +5,6 @@ exclude .gitreview
 include debian/*
 include debian/source/*
 include etc/*
+include etc/neutron/opflex-agent/*
 include rpm/*
 global-exclude *.pyc
--- a/debian/neutron-cisco-apic-host-agent.conf
+++ b/debian/neutron-cisco-apic-host-agent.conf
@@ -15,5 +15,5 @@ end script
 exec start-stop-daemon --start --chuid neutron \
   --exec /usr/bin/neutron-cisco-apic-host-agent -- \
   --config-file=/etc/neutron/neutron.conf \
-  --config-file=/etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini \
+  --config-file=/etc/neutron/opflex-agent/apic_topology_service.ini \
   --log-file=/var/log/neutron/cisco-apic-host-agent.log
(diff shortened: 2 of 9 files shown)


etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini
```diff
@@ -0,0 +1,61 @@
+[ml2_cisco_apic]
```
amittbose: This file perhaps should not be called "neutron/plugins/ml2/ml2_conf_cisco_apic.ini" anymore since it is not used to configure a Neutron ML2 plugin. Something like "neutron/opflex-agent/apic_topology_service.ini" would be more appropriate.

etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini
```diff
@@ -0,0 +1,61 @@
...
+
```
amittbose: This sample config file is misleading - don't think any of these options are applicable to the host-agent. We should put a description and default values for options are applicable in the sample file.

etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini
tbachman: Will change.

etc/neutron/plugins/ml2/ml2_conf_cisco_apic.ini
tbachman: Yeah -- there are just a handful:

https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/apic_topology.py#L54-L64

I'll update this accordingly.

amittbose: You might also consider naming the config file section to something other than "ml2_cisco_apic"

tbachman: How about "apic_host_agent"? None of these are shared with anything.

amittbose: sure

tbachman:
Just pushed an update.

amittbose:
I think you need to fix the following packaging files as well:
1. python-opflex-agent/debian/neutron-cisco-apic-host-agent.conf
2. python-opflex-agent/rpm/neutron-cisco-apic-host-agent.service
3. python-opflex-agent/debian/neutron-opflex-agent.upstart (actually add a reference to the ini file)

tbachman:
Yes (good catch).

tbachman:
Just pushed an update addressing this.

PR #100, opened 2018-02-28
Host agents shouldn't fanout link updates

ivar-lazzaro:
Only one server should serve the RPC call at a time


ivar-lazzaro:
Fanout cast in oslo messaging makes sure that every subscriber reads and processes a specific message (pretty much like broadcast). I don't think this is desirable for LLDP topology service, especially considering that hosts will send updates every minute.
When multiple controllers get these events at the same time, they start fighting each other over the DB transaction and this degrades performance and causes inconsistencies.

tbachman:
Can you also create a similar patch against apic-ml2-driver (master)?

PR #107, opened 2018-03-07
For debian install, removed plugin.ini from unit file

mandeepdhami:
For debian install, removed plugin.ini from unit file

### DIFF
--- a/debian/neutron-opflex-agent.service
+++ b/debian/neutron-opflex-agent.service
@@ -7,7 +7,7 @@ Before=openstack-nova-compute.service
 [Service]
 Type=simple
 User=neutron
-ExecStart=/usr/bin/neutron-opflex-agent --config-file=/etc/neutron/neutron.conf --log-file=/var/log/neutron/opflex-agent.log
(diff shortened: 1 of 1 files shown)


debian/neutron-opflex-agent.service
```diff
@@ -7,7 +7,7 @@ Before=openstack-nova-compute.service
 [Service]
 Type=simple
 User=neutron
-ExecStart=/usr/bin/neutron-opflex-agent --config-file=/etc/neutron/neutron.conf --config-file=/etc/neutron/plugin.ini --log-file=/var/log/neutron/opflex-agent.log
+ExecStart=/usr/bin/neutron-opflex-agent --config-file=/etc/neutron/neutron.conf --log-file=/var/log/neutron/opflex-agent.log
```
amittbose: neutron-opflex-agent reads the config file created for *openvswitch* L2 agent. This other config file is typically located at /etc/neutron/plugins/ml2/openvswitch_agent.ini, and is expected to be loaded by plugin.ini. So, in the least, I think you need to replace plugin.ini with the openvswitch agent's config file.

mandeepdhami: OK, I will add that and update the pull requests.

PR #110, opened 2018-03-08
Add DHCP mac to EP file

tbachman:
When using optimized DHCP, the virtual DHCP server needs to
know the MAC as well as the IP address of the server that
it's virtualizing or proxying.


tbachman:
I wondered this as well when I looked at the "server-ip" parameter. The agent right now only accepts one IP, which is why I was looking to use a single MAC.

mandeepdhami:
The port can be only on a specific subnet (if it has an IP address), we should choose the DHCP for that subnet.

tbachman:
This also needs a respin to match the changes in the plugin -- hold off on merge.

PR #111, opened 2018-03-20
Implement force send on topology agent RPC

ivar-lazzaro:
Although we already had a force_send count in place, adding
an explicit force keyword and making this count less aggressive would
work better with the corresponding server side change.

### DIFF
--- a/opflexagent/apic_topology.py
+++ b/opflexagent/apic_topology.py
@@ -43,7 +43,7 @@ ACI_PORT_LOCAL_FORMAT = 'Eth(\d+)/(\d+(\/\d+)*)'
 ACI_VPCPORT_DESCR_FORMAT = ('topology/pod-(\d+)/protpaths-(\d+)-(\d+)/pathep-'
                             '\[(.*)\]')
 
-AGENT_FORCE_UPDATE_COUNT = 5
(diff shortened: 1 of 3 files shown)


opflexagent/apic_topology.py
```diff
@@ -146,11 +145,14 @@ def _check_for_new_peers(self, context):
...
                 if (interface not in curr_peers or
                         curr_peers[interface] != peer or
                         force_send):
                     LOG.debug('reporting new peer: %s', peer)
-                    self.service_agent.update_link(context, *peer)
+                    self.service_agent.update_link(context, *peer,
+                                                   force=force_send)
+                    self.count_current = 0
                 if interface in curr_peers:
                     curr_peers.pop(interface)
```
kentwu1: do we have to set count_current to 0 also after the update_link() call at line#163? I saw its getting udpated at line#148 so.

tbachman:
Are there server side changes yet? Should that be merged first?

ivar-lazzaro:
Server side change is under review here: https://review.openstack.org/#/c/554414/
I guess we could wait for reviews to approve both these changes before merging them, but doesn't matter which goes first.

PR #113, opened 2018-03-30
request_* RPC API should be of type cast

ivar-lazzaro:
Set fanout to False to make sure only one server at a time serves
the call.

### DIFF
--- a/opflexagent/rpc.py
+++ b/opflexagent/rpc.py
@@ -112,8 +112,8 @@ class GBPServerRpcApi(object):
                                  host=None):
         # Request is a tuple with the device_id as first element, and the
         # request ID as second element
-        cctxt = self.client.prepare(version=self.GBP_RPC_VERSION)
-        cctxt.call(context, 'request_endpoint_details', agent_id=agent_id,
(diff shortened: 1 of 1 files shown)


ivar-lazzaro:
The previous change was made to prevent all the servers from being clogged by the agents' requests. However, by using call instead of cast we also made the RPC blocking, which is not how the agent was designed.
By setting fanout to False, we make sure that only one server at a time can serve the RPC in an asynchronous and round robin fashion

tbachman:
Thanks -- but was hoping we could put that in the commit message?

snaiksat:
So my memory is a little bit vague as to why I chose to use call instead of cast here. There are two possibilities, I either tried setting fanout=False and it didn't work, hence resorted to using call, or I simply didn't recognize the point that we were trying to get the response back asynchronously. But, assuming that setting fanout=False does work correctly, the fix in this patch is definitely seems to be the right solution in the context of what we seem to want to achieve.

tbachman:
Again, all good -- just would love to have this captured in the commit message, since the PR comments don't show up in the git history.

ivar-lazzaro:
Updated commit message

PR #126, opened 2018-04-18
Make topological calls synchronous

ivar-lazzaro:
We need a mechanism to know whether the server was able to update
the host link table. Since such operations are usually quick unless
a host interface has moved somewhere, using blocking calls
will give the agent a way to know whether a link update should be
retried.


ivar-lazzaro:
do not merge yet

ivar-lazzaro:
now we can merge

PR #127, opened 2018-04-18
Load generator

ivar-lazzaro:
Do not merge yet, but leave feedback

### DIFF
--- a/opflexagent/tools/load_generator.py
+++ b/opflexagent/tools/load_generator.py
@@ -1,0 +1,328 @@
+# Copyright (c) 2018 Cisco Systems Inc.
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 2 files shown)


tbachman:
May not need this -- have a look at:

https://github.com/noironetworks/python-opflex-agent/pull/130

ivar-lazzaro:
Yeah that was my initial idea too. Much cleaner and more useful. Will it be possible to run multiple agents in the same host to generate load with that model though?  We might want to add a BridgeManager which polls a file similarly to what this does.

tbachman:
I've tested this using containers, so that it's easily scaled (though so far I've only created 2 containers on a given host). I think the BridgeManager idea is good -- I tried creating a version of the Nova FakeDriver that still does VIF plugging, but my first attempt didn't work, so I'm not yet sure how we'll handle that just yet.

PR #130, opened 2018-04-20
Add fake agent for load testing

tbachman:
This adds a version of the bridge manager that can be used
for load testing. The FakeManager class strictly uses the
update and delete port notifications to generate, update,
and delete EP files. It behaves as the current agent, using
RPCs to get GBP details from the plugin. Because it doesn't
depend on ports being added or removed from an OVS bridge,
it can be used with the FakeDriver in Nova to create
virtual hypervisors, so that laod testing for the network
pieces can be performed, without needing the resources
for compute (storage, CPUs, etc.).

### DIFF
--- a/opflexagent/config.py
+++ b/opflexagent/config.py
@@ -57,6 +57,10 @@ gbp_opts = [
     cfg.StrOpt('fabric_bridge', default='br-fabric',
                help=_("The name of the bridge which connects to the ACI "
                       "fabric")),
+    cfg.StrOpt('bridge_manager',
+               default='ovs',
+               help=_("The class to use for OVS bridge management. "
+                      "Options are: 'ovs' (default), and 'fake'.")),
 ]
 
 cfg.CONF.register_opts(gbp_opts, "OPFLEX")
--- a/opflexagent/opflex_notify.py
+++ b/opflexagent/opflex_notify.py
@@ -147,6 +147,12 @@ class OpflexNotifyAgent(object):
     def run(self):
         """Infinite loop which catches all exception, exits on ^C"""
 
+        # Don't bother running if we don't have a socket
+        if not self.sockname:
+            LOG.warning("Notification socket not set, "
+                        "notifications will not be sent")
+            return
+
         while True:
             try:
                 client = self._connect()
--- a/opflexagent/utils/bridge_managers/ovs_lib.py
+++ b/opflexagent/utils/bridge_managers/ovs_lib.py
@@ -47,3 +47,9 @@ class OVSBridge(ovs_lib.OVSBridge):
                     yield txn
                 finally:
                     self._transaction = None
+
+
+class FakeOVSBridge(OVSBridge):
+
+    def get_vif_port_by_id(self, port_id):
+        return ovs_lib.VifPort(port_id, None, port_id, None, self)
--- a/opflexagent/gbp_ovs_agent.py
+++ b/opflexagent/gbp_ovs_agent.py
@@ -33,6 +33,7 @@ from neutron_lib import constants as n_constants
 from neutron_lib import context
 from neutron_lib import exceptions
 from neutron_lib.utils import helpers
+from neutron_lib.utils import runtime
 from oslo_config import cfg
 from oslo_log import log as logging
 from oslo_service import loopingcall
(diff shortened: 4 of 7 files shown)


opflexagent/config.py
```diff
@@ -57,6 +57,15 @@
     cfg.StrOpt('fabric_bridge', default='br-fabric',
                help=_("The name of the bridge which connects to the ACI "
                       "fabric")),
+    cfg.StrOpt('bridge_manager',
```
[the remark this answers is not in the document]
tbachman: Either name works for me -- I picked bridge_manager only because that's the base class that these inherit from. However, I think it makes sense to keep FakeManager in the same file as OvsManager, since it's just a specialization of the OvsManager class.

opflexagent/utils/bridge_managers/ovs_lib.py
```diff
@@ -47,3 +51,10 @@ def ovsdb_transaction(self):
                     yield txn
                 finally:
                     self._transaction = None
+
```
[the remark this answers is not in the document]
tbachman: For this use case, we want it to look like it's a real bridge, so that we can exercise as much of the networking code path as possible. However, since we're faking the compute side, which is what creates these interfaces, we just create the objects in order to ensure that we get basically the same behavior. 

I'll have a look at that pull-request and do a review -- thanks!

opflexagent/config.py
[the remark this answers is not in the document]
shastrinator: Ok, please ignore my comments, I agree with you. I incorrectly assumed what fake meant here, .  I am assuming nova is just adding the ports to ovsdb directly in this case.  I see that the specific bridge manager class does not have to be imported in gbp_agent as well. I will use this method as well. Thanks!

opflexagent/config.py
```diff
@@ -57,6 +57,15 @@
...
+               default=('opflexagent.utils.bridge_managers.'
+                        'ovs_manager.OvsManager'),
```
ivar-lazzaro: Since we are making a pluggable manager, should we use entry points instead of paths?

opflexagent/opflex_notify.py
```diff
@@ -177,7 +177,9 @@ def __init__(self):
 
         def run(self):
             self.agent = OpflexNotifyAgent()
-            self.agent.run()
+            # Don't bother running if we don't have a socket
+            if self.agent.sockname:
+                self.agent.run()
```
ivar-lazzaro: maybe the run method should be a noop if there's no sockname (logging something meanwhile) instead of leaving the check to the caller.

opflexagent/utils/bridge_managers/ovs_lib.py
```diff
@@ -47,3 +51,10 @@ def ovsdb_transaction(self):
...
+
+class FakeOVSBridge(OVSBridge):
+
+    def get_vif_port_by_id(self, port_id):
+        return ovs_lib.VifPort(port_id, None, port_id,
+            net.get_random_mac(cfg.CONF.base_mac.split(':')), self)
```
ivar-lazzaro: every call to the same port ID will generate a new random mac. Is that what we want? Should we keep the generated value in memory if the same port is called twice?

opflexagent/config.py
tbachman: Good suggestion -- will fix.

opflexagent/utils/bridge_managers/ovs_lib.py
tbachman: It ends up the MAC generated here won't get used. The fake manager only gets triggered by port update events, which means the MAC is already assigned at the plugin. The places where the VIF port object's MAC is referenced are also checking the details first:

https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/utils/ep_managers/endpoint_file_manager.py#L114

https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/utils/ep_managers/endpoint_file_manager.py#L156

https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/utils/ep_managers/endpoint_file_manager.py#L262

tbachman: I guess I should try just passing in None as the MAC -- that should work as well.

opflexagent/opflex_notify.py
tbachman: Sounds good -- will fix.

tbachman:
Pushed update addressing reviewer comments.

PR #147, opened 2018-06-07
VPP agent changes:

shastrinator:
1. VPP dhcp driver and host-interface changes
2. Restartability for gbp-vpp-agent
3. More interface settings: state/mtu
4. Adopting vpplib in opflexagent path
5. vpplib connection fixes and cleanup


shastrinator:
Committing to newton

PR #148, opened 2018-06-12
Fix entry point

tbachman:
Commit f811b830bc38be43e566ab34d5c38a2266c1fe0f changed
the module name, but not the entry point. This fixes the
entry point.

### DIFF
--- a/opflexagent/utils/bridge_managers/vpp_manager.py
+++ b/opflexagent/utils/bridge_managers/vpp_manager.py
@@ -13,6 +13,7 @@
 from neutron.agent.linux import ip_lib
 from neutron.plugins.ml2.drivers.openvswitch.agent.common import constants
 from neutron_lib import constants as lib_constants
+from opflexagent import constants as ofcst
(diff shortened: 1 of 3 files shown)


tbachman:
Good suggestion -- pushed new update, with this change.

opflexagent/utils/bridge_managers/vpp_manager.py
```diff
@@ -33,6 +34,7 @@ def __init__(self):
     def initialize(self, host, conf, agent_state):
         self.int_br_device_count = 0
         vpp_config = conf.VPP
+        agent_state['agent_type'] = ofcst.AGENT_TYPE_OPFLEX_OVS
```
amittbose: this should be AGENT_TYPE_OPFLEX_VPP ?

tbachman: Definitely :-(. Will respin, and thanks for catching!

tbachman:
Pushed new version, addressing reviewer comments (and my own shortcomings ;-) ).

PR #161, opened 2018-09-11
Update the versions for Queens

kentwu1:


snaiksat:
We need to make sure none of the nightly jobs and/or automation is pointing to the master branch before we merge this commit. Also need to  create the queens branch post merging this commit. I will leave it to Thomas to merge this.

kentwu1:
Actually the queens branch was already there. Its just that I missed updating the versions in test-requirement file when I did it first time. So After merging this patch, I will also back-port it to the queens branch.

PR #176, opened 2018-10-17
SNAT conn track logging feature

kentwu1:
1. We will launch/terminate the opflex-conn-track process
thru supervisord when snat netns is being created/deleted.

2. The real command is in opflex_conn_track.py where it
pipes the conntrack output to syslog.


tbachman:
LGTM. We should probably add a config variable that allows people to turn this off, but that can be handled in a separate patch.

PR #230, opened 2019-01-30
Stable/queens

shastrinator:
Fixes from testing vhostuser datapath
    
vhostuser configuration should go to configurations.
As part of the agent state sync, only configurations section is updated.
The names used are also standard and being used for upstream dpdk.
Using these at the agent_state level is incompatible with neutron.


shastrinator:
Incorrect. Please ignore..

tbachman:
A couple of things:
1) For stable branch backports, our policy has been to allow self-merges
2) It looks like the cherry-pick to the stable branch might be wrong -- it seems to include a ton of commits. How did you do the cherry-pick?

shastrinator:
Sorry I did not need this pull request. I had to ignore it, so I just closed the request.. I think git is assuming that I tried to push a stable branch( like a feature branch) on to master, which was not the intention and that is why it included all the commits. The one that got merged has the right cammit.  Is it ok now or did I still do a mistake?

PR #232, opened 2019-02-06
Ensure virtual-ip has CIDR or IPs, not both

tbachman:
The plugin now supports passing allowed-address-pair CIDRs instead
of just /32 host addresses. This resulted in both the CIDR and any
owned addresses in the CIDR to be addeed to the virtual-ip list in
the EP file. While this doesn't break anything, it results in other
agents to log warning messages, as there is no need for both (the
virtual-ip list is used to catch gratuitous ARP packets to the
controller, so it can request exclusive ownership of the IP from
the plugin; since the CIDR already covers the range, any owned /32
IP in the CIDR is redundant). This patch ensures that either the
CIDR or a /32 address will appear in the virtual-ip list, but not
both. Note that CIDRs will never be sent with the "active" flag
set to true in the allowed_address_pairs attribute, so they will
never be added to the ip or anycast_return_ip lists.


opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -342,13 +342,38 @@ def _mapping_to_file(self, port, mapping, fixed_ips):
...
                                         mapping_dict)
         ips_aap = []
-        for aap in mapping.get('allowed_address_pairs', []):
+
+        def filter_cidr_aaps(aaps):
+            # Remove any CIDRs in the AAP list
+            cidr_aaps = [aaps.pop(index)
+                         for index in range(len(aaps))[::-1]
+                         if netaddr.IPNetwork(
+                             aaps[index]['ip_address']).prefixlen != 32]
```
kentwu1: will need to also consider ipv6 case per Bob's comment on the corresponding GBP patch.

opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -342,13 +342,38 @@ def _mapping_to_file(self, port, mapping, fixed_ips):
...
+        aaps = mapping.get('allowed_address_pairs', [])
+        removed_aaps = filter_cidr_aaps(aaps)
+        for aap in aaps:
             if aap.get('ip_address'):
                 virtual_ips.append(
                     {'ip': aap['ip_address'],
                      'mac': aap.get('mac_address', mac)})
                 if aap.get('active'):
                     ips_aap.append(aap['ip_address'])
+        # The removed /32 IPs still need to appear in the other lists
```
kentwu1: ditto

opflexagent/utils/ep_managers/endpoint_file_manager.py
tbachman: Good point -- will fix.

opflexagent/utils/ep_managers/endpoint_file_manager.py
tbachman: Ditto :-)

tbachman:
Pushed an updated commit, addressing reviewer comments.

PR #237, opened 2019-02-26
Fixes from VPP datapath bring-up

shastrinator:
Abstract direct accesses to integration bridge
Fix VPP package paths since package was relocated
Fix return value in VPP api
Fix unavailable constant in Ocata

### DIFF
--- a/opflexagent/snat_iptables_manager.py
+++ b/opflexagent/snat_iptables_manager.py
@@ -27,13 +27,13 @@ LOG = logging.getLogger(__name__)
 class SnatIptablesManager(object):
     IFACE_PREFIX = 'of-'
 
-    def __init__(self, int_br):
-        self.int_br = int_br
+    def __init__(self, bridge_manager):
+        self.bridge_manager = bridge_manager
         self.snat_conn_track_handler = (
-            as_metadata_manager.SnatConnTrackHandler())
+            as_metadata_manager.SnatConnTrackHandler(bridge_manager))
 
     def _cleanup(self, if_name, ns_name):
-        self.int_br.delete_port(if_name)
+        self.bridge_manager.delete_port(if_name)
         ip_wrapper_root = ip_lib.IPWrapper()
         if ip_wrapper_root.netns.exists(ns_name):
             # This will in turn shut down conn_track process thru supervisord
@@ -42,7 +42,7 @@ class SnatIptablesManager(object):
             ip_wrapper_root.netns.delete(ns_name)
 
     def _add_port_and_netns(self, if_name, ns_name, if_mac=None, mtu=None):
-        self.int_br.add_port(if_name, ('type', 'internal'))
+        self.bridge_manager.add_port(if_name, ('type', 'internal'))
         ip_wrapper_root = ip_lib.IPWrapper()
         if_dev = ip_wrapper_root.device(if_name)
         if if_mac:
@@ -155,7 +155,7 @@ class SnatIptablesManager(object):
 
     def cleanup_snat_all(self, exclude_es=[]):
         exclude_ports = set([self._get_hash_for_es(e) for e in exclude_es])
-        ports = [p for p in self.int_br.get_port_name_list()
+        ports = [p for p in self.bridge_manager.get_port_name_list()
                  if p.startswith(self.IFACE_PREFIX) and p not in exclude_ports]
         for ifn in ports:
             self._cleanup(ifn, ifn)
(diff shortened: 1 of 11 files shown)


opflexagent/vif_plug_vpp/linux_net.py
```diff
@@ -42,7 +41,7 @@ def _create_vpp_vif(dev, iface_id, mac, mtu,
                         1, mac_address, iface_id)
     LOG.debug("sw_if_index:{}".format(sw_if_index))
     if mtu:
-        vapi.set_interface_mtu(sw_if_index, mtu)
+        vapi.set_interface_mtu(sw_if_index, [mtu, 0, 0, 0])
```
kentwu1: not sure what the extra 3 0's mean here.....

opflexagent/vpplib/VPPApi.py
```diff
@@ -49,8 +49,7 @@ def __init__(self, client_name, LOG):
         self.connected = False
 
     def __enter__(self):
-        self.vppp = VppPapiProvider(self.client_name, self.LOG,
-            None, self.read_timeout)
+        self.vppp = VppPapiProvider(self.client_name, None, self.read_timeout)
```
kentwu1: Wondering how this error can get away with it before.....Don't we have a UT in place to exercise this code path then?

opflexagent/vpp_plug_firewall.py
```diff
@@ -28,8 +28,7 @@ def vpp_nova_to_osvif_vif_vhostuser(vif):
         except NotImplementedError as e:
             if vif['details'].get("vhostuser_vpp_plug", False):
                 profile = objects.vif.VIFPortProfileOpenVSwitch(
-                    interface_id=vif.get('ovs_interfaceid') or vif['id'],
-                    datapath_type=model.VIF_DETAILS_OVS_DATAPATH_NETDEV)
+                    interface_id=vif.get('ovs_interfaceid') or vif['id'])
```
kentwu1: not that familiar with the code but why we don't need to pass in the datapath_type parameter now?

opflexagent/vpplib/VPPApi.py
shastrinator: I have used mock APIs at this level in the UT intentionally. We were not sure about whether we would release this library or the VPP team. I had all these changes locally and missed committing  the last time around, which was why I retested this path.

opflexagent/vif_plug_vpp/linux_net.py
shastrinator: 4 values are for [all packets, L3 packets, IPv4 packets, MPLS packets], first value applies to all unless overriden.

opflexagent/vpp_plug_firewall.py
shastrinator: This function derives from nova/nova/network/os_vif_util.py _nova_to_osvif_vif_vhostuser. This argument was added post ocata. I had to make this change for testing in ocata. For the master branch, I can have this argument still, I will add it back. Thanks.

opflexagent/snat_iptables_manager.py
```diff
@@ -27,13 +27,13 @@
...
-    def __init__(self, int_br):
-        self.int_br = int_br
+    def __init__(self, bridge_manager):
+        self.bridge_manager = bridge_manager
         self.snat_conn_track_handler = (
-            as_metadata_manager.SnatConnTrackHandler())
+            as_metadata_manager.SnatConnTrackHandler(bridge_manager))
 
     def _cleanup(self, if_name, ns_name):
-        self.int_br.delete_port(if_name)
```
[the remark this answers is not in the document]
shastrinator: Sure, I was misled by the same name. I will change them to use fabric_br. No UT failed though. Thanks for catching this.

opflexagent/utils/bridge_managers/ovs_manager.py
```diff
@@ -170,6 +170,18 @@ def get_vif_port_by_id(self, id):
     def handle_removed_eps(self, em, removed_eps):
         return removed_eps
 
+    def plug_metadata_port(self, dst_shell, port):
+        dst_shell("ovs-vsctl add-port %s %s" % (self.fabric_br.br_name, port))
+
+    def delete_port(self, port):
+        self.int_br.delete_port(port)
```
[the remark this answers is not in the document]
shastrinator: Sure

kentwu1:
It looks ok to me now but have you run some sanity test on a real opflex fab to make sure its working as expected? Thx.

tbachman:
The pep8 tox job runs the DB migration, and that worked.

PR #243, opened 2019-04-05
Add VRF notify RPC

tbachman:
This adds an RPC notification for VRF updates. The notification
causes the agent to make its request_vrf_details RPC, in order
to retrieve any changes to the subnets located in this VRF.


tbachman:
Minor update to the patch - needed to add the topic for the agent to listen to.

PR #261, opened 2019-05-08
Fix AsMetadata Manager init

shastrinator:
In cases where AsMetaDataManager is not started from opflex_agent,
we can directly get the correct bridge_manager class to load
and create the metadata port on the bridge from config_opts.
Bridge is assumed to be initialized before metadata manager is invoked,
as before.

Signed-off-by: Kiran Shastri <shastrinator@gmail.com>

### DIFF
--- a/opflexagent/as_metadata_manager.py
+++ b/opflexagent/as_metadata_manager.py
@@ -29,6 +29,7 @@ from neutron.common import utils
 from neutron.conf.agent import common as config
 from neutron.plugins.ml2.drivers.openvswitch.agent.common import (  # noqa
     config as ovs_config)
+from opflexagent.utils import utils as opflexagent_utils
 from oslo_config import cfg
 from oslo_log import log as logging
 from oslo_serialization import jsonutils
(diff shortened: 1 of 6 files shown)


opflexagent/as_metadata_manager.py
```diff
@@ -82,6 +85,22 @@
 STATE_FILENAME_NETS = STATE_FILE_NAME_FORMAT % STATE_INSTANCE_NETWORKS
 
 
+def get_bridge_manager(conf):
```
tbachman: We should move this into a shared utils directory, and have the gbp_agent use the same method. Maybe create a new opflexagent/utils/utils.py file for this?

opflexagent/as_metadata_manager.py
```diff
@@ -391,7 +410,8 @@ def process(self, files):
 class StateWatcher(FileWatcher):
     def __init__(self):
         root_helper = cfg.CONF.AGENT.root_helper
-        self.mgr = AsMetadataManager(LOG, root_helper)
+        bridge_manager = get_bridge_manager(cfg.CONF.OPFLEX)
```
tbachman: I don't think this will work - this is just the class, and the AsMetadataManager is expecting an instance. Also, the plug method requires that the initialize method on the bridge_manager has been called. I think we need a different approach for all of this.

shastrinator: Have fixed this.

opflexagent/as_metadata_manager.py
shastrinator: Accepted. Have changed this

PR #279, opened 2019-07-29
Fix for conntrack logging

tbachman:
The SNAT conntrack logging processes weren't getting restarted
if killed. This was because the return code of the conntrack
command wasn't being propagated, so the opflex-conn-track command
returned an exit status of 0, indicating normal termination.

### DIFF
--- a/opflexagent/opflex_conn_track.py
+++ b/opflexagent/opflex_conn_track.py
@@ -45,9 +45,7 @@ def main():
                "-p %s.%s -t opflex-conn-track") % (
                    sys.argv[1], sys.argv[2], sys.argv[3])
     LOG.debug("conn_track command: %s" % command)
-    sh(command)
-
-    return
+    return sh(command)
 
 
 if __name__ == "__main__":


opflexagent/as_metadata_manager.py
```diff
@@ -560,6 +560,7 @@ def conn_track_del(self, netns):
     def conn_track_config(self, netns):
         snatstr = "\n".join([
             "[program:opflex-conn-track-%s]" % netns,
+            "process_name=opflex-conn-track-%s" % netns,
```
kentwu1: do we have to worry about this guy also then since I don't see process_name attribute there either?
https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/as_metadata_manager.py#L510

tbachman: The diagnosis was incorrect - pushed an update with the correct fix.

opflexagent/opflex_conn_track.py
```diff
@@ -45,9 +45,9 @@ def main():
                "-p %s.%s -t opflex-conn-track") % (
                    sys.argv[1], sys.argv[2], sys.argv[3])
     LOG.debug("conn_track command: %s" % command)
-    sh(command)
+    ret = sh(command)
```
kentwu1: nit: might be more straightforward and clean to just do "return sh(command)" then?

tbachman: pushed update, addressing this comment.

PR #284, opened 2019-08-06
Fix SNAT conntrack logging cleanup

tbachman:
The current logging for connection tracker events for the SNAT
linux namespaces does not get cleaned up correctly in all cases.
One of the problems was that two processes were spawned in the
same shell command, leaving one process untrackable. This patch
usees the Popen method of subprocess in order to explicitly spawn
and track both child processes, and adds a parent class to manage
the cleanup and return codes in all cases, in order to allow the
supervisord to restart dead processes when needed.

### DIFF
--- a/opflexagent/opflex_conn_track.py
+++ b/opflexagent/opflex_conn_track.py
@@ -11,8 +11,10 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+import signal
 import subprocess
 import sys
+import time
 
 from neutron.common import config
 from neutron.common import utils as comm_utils
@@ -22,16 +24,77 @@ from oslo_log import log
 LOG = log.getLogger(__name__)
 
 
-def sh(cmd):
-    LOG.debug("conn_track: Running command: %s" % cmd)
-    ret = ''
-    try:
-        ret = subprocess.check_output(
-            cmd, stderr=subprocess.STDOUT, shell=True)
-    except Exception as e:
-        LOG.error("In running command: %s: %s" % (cmd, str(e)))
-    LOG.debug("conn_track: Command output: %s" % ret)
-    return ret
+class SnatConntrackLogger(object):
+    """Opflex SNAT Connection Tracking Logger
+
+    This class dumps the output of connection tracking events
+    for a Linux network namespace used to perform SNAT on hosts
+    running the neutron-opflex-agent. This class is created in
+    order to spawn two processes per namespace: one to run a
+    connection tracker in the Linux network namespace, and the
+    other to collect the standard output and dump it into syslog,
+    using the log level and facility set in a configuration file.
+    The process for this is managed using supervisord, which in
+    version 4.0 or later supports redirecting stdout and stderr
+    to syslog. However, older Linux distributions ship with
+    supervisord versions older than this, so the logger tool
+    must be used for this redirection. While supervisord can
+    handle the lifecycle of the top level process, this class
+    neededs manage the lifecycle of the child prcoesses that
+    are spawned (i.e. conntrack and logger), ensuring that
+    the parent exit code is set correctly so that supervisord
+    can handle any respawning, if needed.
+
+    REVISIT: Should have some sort of pooling to avoid exhaustion
+    """
+    def __init__(self, conntrack_cmd, logger_cmd):
+        self.ret = None
+        self.p1 = self.p2 = None
+        try:
+            # Conntrack process redirects stderr to stdout,
+            # returned Popen has file object for stdout
+            self.p1 = subprocess.Popen(
+                conntrack_cmd, stdout=subprocess.PIPE,
+                stderr=subprocess.STDOUT, close_fds=True, shell=True)
+            # Logger uses stdout member from conntrack process for its stdin
+            self.p2 = subprocess.Popen(
+                logger_cmd, stdin=self.p1.stdout, close_fds=True, shell=True)
+        except Exception as e:
+            LOG.error("In running commands: %s: %s" %
+                      (conntrack_cmd + logger_cmd, str(e)))
+
+    def _kill_child_procs(self):
+        for proc in [self.p1, self.p2]:
+            LOG.debug("conn_track: proc: %s" % proc.returncode)
+            try:
+                proc.terminate()
+            except Exception:
+                pass
+
+    def terminate(self, signum, frame):
+        self._kill_child_procs()
+        LOG.debug("conn_track: returning %s" % -signum)
+        sys.exit(-signum)
+
+    def wait(self):
+        # REVISIT: We probably should consider whether we
+        #          should be checking for other signals,
+        #          such as SIGPIPE. The assumption is that
+        #          anything uncaught could result in a child
+        #          process not getting terminated, and left
+        #          as a zombie process
+        signal.signal(signal.SIGINT, self.terminate)
+        signal.signal(signal.SIGTERM, self.terminate)
+        while (self.p1 and self.p2) and (self.p1.poll() is None and
+                                         self.p2.poll() is None):
+            time.sleep(1)
+
+        self._kill_child_procs()
+        # NOTE: The assumption here is that this process is always
+        #       being managed by supervisord. In this case, we always
+        #       return a non-zero RC, as we want supervisord to restart
+        #       us whenever we die, regardless of the reason.
+        return -signal.SIGTERM
 
 
 # This program takes 3 parameters,
@@ -41,12 +104,16 @@ def sh(cmd):
 def main():
     config.setup_logging()
     comm_utils.log_opt_values(LOG)
-    command = ("ip netns exec %s conntrack -E -o timestamp 2>&1 | logger "
-               "-p %s.%s -t opflex-conn-track") % (
-                   sys.argv[1], sys.argv[2], sys.argv[3])
-    LOG.debug("conn_track command: %s" % command)
-    return sh(command)
+    cmd1 = ("ip netns exec %s conntrack -E -o timestamp") % (sys.argv[1])
+    cmd2 = ("logger -p %s.%s -t opflex-conn-track") % (sys.argv[2],
+                                                       sys.argv[3])
+    LOG.debug("conn_track command: %s" % cmd1)
+    LOG.debug("logger command: %s" % cmd2)
+    wrapper = SnatConntrackLogger(cmd1, cmd2)
+    return wrapper.wait()
 
 
 if __name__ == "__main__":
-    main()
+    ret = main()
+    LOG.debug("conn_track: returning %s" % ret)
+    sys.exit(ret)


opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+        for proc in [self.p1, self.p2]:
+            LOG.debug("conn_track: proc: %s" % proc.returncode)
+            try:
+                proc.terminate()
+            except Exception:
+                pass
+
+    def terminate(self, signum, frame):
+        if signum is not None:
+            self._kill_child_procs(save_rc=False)
```
kentwu1: not sure why we don't care about the return code if this program is getting killed by SIGINT and SIGTERM.....

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+    def wait(self):
+        signal.signal(signal.SIGINT, self.terminate)
+        signal.signal(signal.SIGTERM, self.terminate)
+        while (self.p1 and self.p2) and (self.p1.poll() is None and
+                                         self.p2.poll() is None):
+            time.sleep(1)
+
+        self._kill_child_procs()
+        if self.ret is None:
+            self.ret = ''
```
kentwu1: will this '' make superviord to restart the conn_track process then?

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+            if self.p1 and self.p1.poll() is not None:
+                # It must have died, so set it to nonzero RC or SIGTERM
+                self.ret = self.p1.returncode or -signal.SIGTERM
+            if not self.ret and self.p2 and self.p2.poll() is not None:
+                # It must have died, so set it to nonzero RC or SIGTERM
+                self.ret = self.p2.returncode or -signal.SIGTERM
+        for proc in [self.p1, self.p2]:
+            LOG.debug("conn_track: proc: %s" % proc.returncode)
+            try:
+                proc.terminate()
```
abhis2112: proc.terminate method itself is a try except block that shows the error as OS.ERROR if fails in except block. Do we need here try except block explicitly ?

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+            LOG.error("In running commands: %s: %s" %
+                      (conntrack_cmd + logger_cmd, str(e)))
+
+    def _kill_child_procs(self, save_rc=True):
+        if save_rc:
+            # if poll returns None, it means the process hasn't exited yet
+            # In this case, we're going to kill it, so don't save the RC
+            if self.p1 and self.p1.poll() is not None:
+                # It must have died, so set it to nonzero RC or SIGTERM
+                self.ret = self.p1.returncode or -signal.SIGTERM
```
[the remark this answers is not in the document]
tbachman: Just to be consistent - usually error return codes are negative. The constants defined by signal are positive.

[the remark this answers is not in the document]
tbachman: I wasn't sure - but I wanted to be positive that the child processes were cleaned up. I'll investigate that library to find out how this can happen.

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+                proc.terminate()
+            except Exception:
+                pass
+
+    def terminate(self, signum, frame):
+        if signum is not None:
+            self._kill_child_procs(save_rc=False)
+            sys.exit(0)
+        else:
+            self._kill_child_procs()
```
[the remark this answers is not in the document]
tbachman: I wasn't sure - but I wanted to be positive that the child processes were cleaned up. I'll investigate that library to find out how this can happen.

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
-    ret = ''
-    try:
-        ret = subprocess.check_output(
-            cmd, stderr=subprocess.STDOUT, shell=True)
-    except Exception as e:
-        LOG.error("In running command: %s: %s" % (cmd, str(e)))
-    LOG.debug("conn_track: Command output: %s" % ret)
-    return ret
+class SnatConntrackLogger(object):
+    """Opflex SNAT Connection Traking Logger
```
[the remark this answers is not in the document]
tbachman: Will fix

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+
+    This class dumps the output of connection tracking events
+    for a Linux network namespace used to perform SNAT on hosts
+    running the neutron-opflex-agent. This class is created in
+    order to spawn two processes per namespace: one to run a
+    connection tracker in the Linux network namespace, and the
+    other to collect the standard output and dump it into syslog,
+    using the log level and identity set in a configuration file.
```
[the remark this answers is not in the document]
tbachman: will fix

opflexagent/opflex_conn_track.py
tbachman: Those are valid values - we need to have cases where a program is terminated validly. That's taken care of by supervisord.

opflexagent/opflex_conn_track.py
tbachman: That's the way it was previously, and supervisord handled it.

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+        else:
+            self._kill_child_procs()
+            # what should be the return value here?
+
+    def wait(self):
+        signal.signal(signal.SIGINT, self.terminate)
+        signal.signal(signal.SIGTERM, self.terminate)
+        while (self.p1 and self.p2) and (self.p1.poll() is None and
+                                         self.p2.poll() is None):
+            time.sleep(1)
```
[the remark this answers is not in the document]
tbachman: In this case, we want this to go on forever unless either of the children processes are killed or this process gets a signal. I'm using sleep to minimize CPU use. I'm open to other approaches on this if folks have preferences.

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+    to syslog. However, older Linux distributions ship with
+    supervisord versions older than this, so the logger tool
+    must be used for this redirection. While supervisord can
+    handle the lifecycle of the top level process, this class
+    neededs manage the lifecycle of the child prcoesses that
+    are spawned (i.e. conntrack and logger), ensuring that
+    the parent exit code is set correctly so that supervisord
+    can handle any respawning, if needed.
+
+    REVISIT: Should have some sort of pooling to avoid exhaustion
```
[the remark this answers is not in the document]
tbachman: In this case it is pooling - right now, this creates as many processes as there are external networks. It seems like we could/should bound this with a manager that manages these with pools.

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,82 @@
...
+    """
+    def __init__(self, conntrack_cmd, logger_cmd):
+        self.ret = None
+        self.p1 = self.p2 = None
+        try:
+            # Conntrack process redirects stderr to stdout,
+            # retured Popen has file object for stdout
```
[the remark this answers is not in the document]
tbachman: will fix.

opflexagent/opflex_conn_track.py
tbachman: I'll have to retest this - for some reason, I thought I had to add this, but I'll double-check.

opflexagent/opflex_conn_track.py
```diff
@@ -41,11 +109,13 @@ def sh(cmd):
...
-                   sys.argv[1], sys.argv[2], sys.argv[3])
-    LOG.debug("conn_track command: %s" % command)
-    return sh(command)
+    cmd1 = ("ip netns exec %s conntrack -E -o timestamp") % (sys.argv[1])
+    cmd2 = ("logger -p %s.%s -t opflex-conn-track") % (sys.argv[2],
+                                                       sys.argv[3])
+    LOG.info("conn_track command: %s" % cmd1)
+    LOG.info("logger command: %s" % cmd2)
+    wrapper = SnatConntrackLogger(cmd1, cmd2)
+    return wrapper.wait()
```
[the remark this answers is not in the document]
tbachman: Hmm - I'll look into this. To be honest, I left this from the original patch, but it's definitely worth investigating. If that's the case, then I'll change it.

opflexagent/opflex_conn_track.py
kentwu1: It seems to me that even for this case, its still no harm to save the return code right? If so, we can really consolidate the code paths and make it more straightforward.

opflexagent/opflex_conn_track.py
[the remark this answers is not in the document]
tbachman: Looks like we should always get a signal, so I will remove the else clause.

opflexagent/opflex_conn_track.py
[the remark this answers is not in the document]
tbachman: It ends up this is just a way of ensuring that the code could be called if it were included as a python module, or run independently as main. For example:

import opflex_conn_track as oct

oct.main()

OR

# python opflex_conn_track.py 

I doubt that we'll ever import this module and call its main, so we could probably put this code under that method, but from what I've seen in looking at other python code, this seems to be a common convention that's used (i.e. define main method that is called when __name__ == "__main__"). In fact, we currently don't import anything from it at all, so we could get rid of the __name__ == "__main__" check entirely, but this again the convention seems to be to keep the __name__ check. If folks feel strongly about it, then I think we should both get rid of the __name__ check and just put the stuff under main as module-level statements.

opflexagent/opflex_conn_track.py
tbachman: I'm changing this to always exit with signum.

opflexagent/opflex_conn_track.py
tbachman: I'm thinking that the above is probably incorrect - from this link (see "exitcodes" section):
http://supervisord.org/configuration.html#program-x-section-settings

supervisord is configured to restart the process unless it gets an exit code of 0 or 2 (https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/as_metadata_manager.py#L565). 0 is normal, 2 is SIGINT, which I guess are considered "normal" exit codes. SIGTERM is 15, which is what happens when I kill the the top level process (sys.exit in line 86), and when wait exits due to child process getting killed. I'll change this to drop the empty string.

opflexagent/opflex_conn_track.py
tbachman: I retested, and something here i sneeded - if I don't do that, and the process has already died, I get this when I call terminate:
http://paste.openstack.org/show/755694/

I could also use a call to poll to check, but that could be racy, and using try/except is also a common idiom in python.

tbachman:

opflexagent/opflex_conn_track.py
kentwu1: do we still plan to change self.ret = '' to something else then?

opflexagent/opflex_conn_track.py
```diff
@@ -22,16 +24,79 @@
...
+                stderr=subprocess.STDOUT, close_fds=True, shell=True)
+            # Logger uses stdout member from conntrack process for its stdin
+            self.p2 = subprocess.Popen(
+                logger_cmd, stdin=self.p1.stdout, close_fds=True, shell=True)
+        except Exception as e:
+            LOG.error("In running commands: %s: %s" %
+                      (conntrack_cmd + logger_cmd, str(e)))
+
+    def _kill_child_procs(self, save_rc=True):
+        if save_rc:
```
kentwu1: Wondering is it worth the effort to make this API to always do 'save_rc=True' case regardless so that we can consolidate the code paths. It seems no harm to me even its called by terminate() with save_rc=True.

tbachman: Heh - I forgot to delete the self.ret is None check. Will fix. As for the _kill_child_procs method - we actually want both. In the case where one of the child processes exits, we want to get the return code (i.e. in main). In the case where the parent was terminated, we don't want their return codes.

tbachman:
pushed another iteration

PR #304, opened 2019-11-22
Send port_update notification to bound host only

rkukura:
When a port is updated, there is generally no need to notify opflex agents on hosts other than the one where the port is bound. Therefore, instead of sending a fanout cast, the port_update notification is sent only to the server where the port is currently bound.

We need to ensure that during VM migration, we are not dependent on the old bound host getting a notification after the port has been rebound to the new host.


tbachman:
Please backport through stable branches to stable/newton

PR #317, opened 2019-12-02
Changed config default for parameter nested_domain_uplink_interface f…

irathore:
…rom patch-fabric-ex to patch-fab-ex


tbachman:
Thanks for the patch! Please backport through stable/newton.

PR #318, opened 2019-12-03
Changed config default for param nested_domain_uplink_interface

irathore:
from patch-fabric-ex to patch-fab-ex

(cherry picked from commit 5103a14ff46551031030906a85769ccc5de6245f)


tbachman:
Thanks Ifti! You can self-merge the remaining backports if you want.

PR #324, opened 2019-12-24
Add the active-active-aap flag into the EP file

kentwu1:


opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -388,6 +388,8 @@ def filter_cidr_aaps(aaps):
         if ips or ips_aap:
             mapping_dict['anycast-return-ip'] = sorted(ips + ips_aap)
 
+        if 'active_active_aap' in mapping:
```
tbachman: Couldn't you just do mapping_dict['active-active-aap'] = mapping.get('active_active_aap', False)?

kentwu1: sure I can do that np.

kentwu1: Hmm looks like changing this actually messed up quite a few UTs......That's probably why in this file we can see lots of "if XXX in mapping:" patterns. I would just follow suit then if no objections.

tbachman: That's fine - go ahead and revert to the original patch.

kentwu1:
Pushed a revision to revert it back to the original patch.

tbachman:
Thanks - please post the backports through stable/newton.

PR #325, opened 2020-01-17
Additional fields on EP file for VMs on SVI networks

skandasw:
We now support VM instantiation on an SVI network on hosts
running opflex-agent. To support the associated data path changes
additional fields are introduced in the EP file as:
- ext-svi             : flag to indicate that the network is of
                        type SVI
- ext-encap-id        : indicates the vlan-id
- endpoint-group-name : now contains a unique id for an SVI network
                        so we use the network uuid
These are obtained in response to the get_gbp_details rpc.

### DIFF
--- a/opflexagent/test/test_endpoint_file_manager.py
+++ b/opflexagent/test/test_endpoint_file_manager.py
@@ -1084,3 +1084,23 @@ class TestEndpointFileManager(base.OpflexTestBase):
         # Get rid of the EPG, verify we don't get an exception
         mapping['endpoint_group_name'] = None
         self.manager.declare_endpoint(port_1, mapping)
+
+    def test_svi_port_bound(self):
+        # the SVI related info we expect to see
+        # on get_gbp_details
+        svi_info = {}
+        svi_info['svi'] = True
+        svi_info['svi_vlan'] = 1234
+        svi_info['endpoint_group_name'] = 'svi-net-id'
+
+        mapping = self._get_gbp_details(**svi_info)
+        port = self._port()
+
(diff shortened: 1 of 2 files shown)


skandasw:
Thanks Thomas. Added a UT.

opflexagent/test/test_endpoint_file_manager.py
```diff
@@ -1084,3 +1084,27 @@ def test_bad_mapping(self):
...
+
+    def test_svi_port_bound(self):
+        # the SVI related info we expect to see
+        # on get_gbp_details
+        svi_info = {}
+        svi_info['svi'] = True
+        svi_info['svi_vlan'] = 1234
+        svi_info['endpoint_group_name'] = 'svi-net-id'
+        # knock off some extraneous stuff
+        svi_info['floating_ip'] = []
```
tbachman: Do we not support floating IPs or SNAT with SVI?

skandasw: No its not that, for the purposes of testing this change was keeping things more focussed on the relevant fields. But perhaps it gives the wrong idea ?

tbachman: Just seemed a bit odd that these were explicitly removed. I'd keep them in if it doesn't affect anything.

skandasw: Updated as this is not as relevant to the change.

tbachman:
Thanks! Please backport through the stable branches. I believe you'll only need to go as far as ocata, as I don't think we supported SVI networks in newton, but please double-check.

PR #326, opened 2020-01-25
Add vlan type networks to allowed list for opflex agent

skandasw:
With support for SVI we need to allow vlan type networks.

### DIFF
--- a/opflexagent/test/test_gbp_ovs_agent.py
+++ b/opflexagent/test/test_gbp_ovs_agent.py
@@ -413,15 +413,16 @@ class TestGBPOpflexAgent(base.OpflexTestBase):
         self.agent.bridge_manager.delete_patch_ports.assert_called_with(
             [subports[0].port_id, subports[1].port_id])
 
-    def test_port_bound_to_host(self):
+    def _test_port_bound_to_host(self, net_type):
         mapping = self._get_gbp_details(device='some_device')
+        seg_id = 1234 if net_type is 'vlan' else ''
         port_details = {'device': 'some_device',
                         'admin_state_up': True,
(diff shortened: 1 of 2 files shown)


opflexagent/test/test_gbp_ovs_agent.py
```diff
@@ -413,13 +413,13 @@ def binding_call(context, subports):
...
+    def _test_port_bound_to_host(self, net_type):
         mapping = self._get_gbp_details(device='some_device')
         port_details = {'device': 'some_device',
                         'admin_state_up': True,
                         'port_id': mapping['port_id'],
                         'network_id': 'some-net',
-                        'network_type': 'opflex',
+                        'network_type': net_type,
                         'physical_network': 'phys_net',
                         'segmentation_id': '',
```
tbachman: This is perhaps a nit, but I'm thinking that if we're supporting vlan network types, we might also want to supply a valid segmentation_id (currently an empty string). The UT isn't checking, but it might be good to at least set this value to a valid VLAN VNID, and perhaps check it.

skandasw: Thx Thomas. The intent here was just to make sure that we allow vlan type networks. 
There is another UT which I added: opflexagent.test.test_endpoint_file_manager.TestEndpointFileManager.test_svi_port_bound
which validates the vlan-id.
But let me take a look.

skandasw: Thx Thomas for the clarification - my concern was with the mocking it may not add much value - but yes now it does not read weird.  And yes the test mentioned above will validated the vlan id.

PR #331, opened 2020-01-31
Move VPCMODULE to constants

tbachman:
The VPCMODULE string format is moved to the constants file, so
that it can be shared for use outside of the agent.


skandasw:
Looks good to me.

irathore:
+1

PR #349, opened 2020-03-03
Fix OVS option registration

tbachman:
The upstream OVS configuration options weren't being referenced
at the time of import ofr the OVS bridge manager, which resulted
in exceptions like the following:

oslo_config.cfg.NoSuchOptError: no such option bridge_mappings in group [OVS]

This patch imports the appropriate libraries to ensure that the configuration
options are loaded so that they can be used when the OVS bridge manager
is loaded.


tbachman:
Will be covered by PR #351

PR #350, opened 2020-03-05
python-opflex-agent supports py36

shyam81295:
This patch aims at supporting 'python-opflex-agent' repo with python3.6 compatibility.

1. Deprecation and removal of `with contextlib.nested` for "mocks" issue is fixed by using literally nested `mock.patch`.

2. By mocking `os.path.basename` fixed the issue in mocking `sys.argv` in py36.

3. Fixed common py2 <-> py3 compatibility errors: 

- Replacing `dict.iteritems()` with `dict.items()`.

- Using `open()` instead of `file()`.

- Using `list` when python dict size changes.

- Removal of `Exception.message` in python3

### DIFF
--- a/opflexagent/test/test_agent_types.py
+++ b/opflexagent/test/test_agent_types.py
@@ -38,10 +38,11 @@ class TestGBPOpflexAgentTypes(base.OpflexTestBase):
             'opflexagent.as_metadata_manager.AsMetadataManager')
         metadata_patch = metadata_mgr.start()
         cfg.CONF.set_override('agent_mode', 'opflex', 'OPFLEX')
-        with mock.patch('sys.argv'):
-            gbp_agent.main()
-            self.assertEqual(1, opflex_patch.call_count)
-            self.assertEqual(1, metadata_patch.call_count)
+        with mock.patch('os.path.basename'):
+            with mock.patch('sys.argv'):
+                gbp_agent.main()
+                self.assertEqual(1, opflex_patch.call_count)
+                self.assertEqual(1, metadata_patch.call_count)
         opflex_agent.stop()
         metadata_mgr.stop()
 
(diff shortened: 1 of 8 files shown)


opflexagent/test/test_agent_types.py
```diff
@@ -38,10 +38,11 @@ def test_opflex_agent_mode(self):
             'opflexagent.as_metadata_manager.AsMetadataManager')
         metadata_patch = metadata_mgr.start()
         cfg.CONF.set_override('agent_mode', 'opflex', 'OPFLEX')
-        with mock.patch('sys.argv'):
-            gbp_agent.main()
-            self.assertEqual(1, opflex_patch.call_count)
-            self.assertEqual(1, metadata_patch.call_count)
+        with mock.patch('os.path.basename'):
```
tbachman: This is a general comment on replacing contextlib.nested patterns, which is used throughout this patch (although this line didn't have it originally, this fix for python3 would benefit from this pattern if it were available). In python3, it's possible to just use a comma-delimited list of statements that need a context manager, instead of cascading them like this (see https://books.google.com/books?id=flRFgvAU7N8C&pg=PA370&lpg=PA370&dq=python3+contextlib+nested+replacement&source=bl&ots=pFTexqEi0Q&sig=ACfU3U1eLdrdddgK5PRWQ-dO0ss7ZBe5ZA&hl=en&sa=X&ved=2ahUKEwjrwuq-qYLoAhXJGM0KHfbeBvEQ6AEwCXoECAoQAQ#v=onepage&q=python3%20contextlib%20nested%20replacement&f=false and https://stackoverflow.com/questions/16083791/alternative-to-contextlib-nested-with-variable-number-of-context-managers). I think these alternatives would read better than the cascaded approach.

opflexagent/utils/port_managers/async_port_manager.py
```diff
@@ -101,7 +101,7 @@ def apply_config(self):
             with excutils.save_and_reraise_exception():
                 # The upper layers will trigger the resync
                 LOG.error("Configuration failed on port manager: %s",
-                          e.message)
+                          str(e))
```
tbachman: Are you sure the str() is needed here? I would have thought that the format API handled this for you?

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
```
tbachman: One other thing I should have told you about - we run a Travis CI job with every patch in this repo. That's covered by this file:
https://github.com/noironetworks/python-opflex-agent/blob/master/.travis.yml

We should add python3 tests to the Travis CI job as part of this patch.

opflexagent/utils/port_managers/async_port_manager.py
shyam81295: Yes, str() prints the Exception message for both py2 and py3, and Exception.message was deprecated & removed in python3.

opflexagent/test/test_agent_types.py
niteshk4-bld: we have removed nested Mock and we used ExitStack in place of that.

PR #352, opened 2020-03-07
Modify EP file field name to indicate SVI and vlan type net

skandasw:
Since the workflow for provider vlan nets is identical to SVI nets,
change the field from 'ext-svi' to 'provider-vlan' for clarity.


skandasw:
In addition to review comments - we shd wait on data path and neutron changes to merge. I am thinking to change the key on rpc to perhaps something simpler as provider_vlan to align with the EP file field. But will track with the neutron changes.

shastrinator:
Please wait for this commit:
https://git.opendaylight.org/gerrit/c/opflex/+/88367

PR #354, opened 2020-04-01
Updates to handling for EP file with VMs on VLAN (non SVI) nets

skandasw:
- Support VLAN (non SVI) nets.
- Similar to SVI except on how EPG name is handled. For SVI,
  it is overloaded with a unique id as received on the RPC.
  With the VLAN (non SVI) case the actual EPG name is used.
- Added config option to track physnet mappings for VLAN type
  nets.
- Some cosmetic changes to fields for better readability.

### DIFF
--- a/opflexagent/utils/ep_managers/endpoint_file_manager.py
+++ b/opflexagent/utils/ep_managers/endpoint_file_manager.py
@@ -325,6 +325,9 @@ class EndpointFileManager(endpoint_manager_base.EndpointManagerBase):
             'neutron-metadata-optimization':
                 mapping['enable_metadata_optimization'],
         }
+        if port.network_type == n_constants.TYPE_VLAN:
+            mapping_dict['provider-vlan'] = True
+            mapping_dict['ext-encap-id'] = port.segmentation_id
         if mapping.get('svi'):
             # VM on SVI type network, in addition to the flag and
             # vlan-id, epg is set to a unique id so using the network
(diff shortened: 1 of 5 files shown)


skandasw:
Validated with basic ping across computes on a FAB  with VMs on vlan net as well Opflex nets.

opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -325,15 +325,16 @@ def _mapping_to_file(self, port, mapping, fixed_ips):
             'neutron-metadata-optimization':
                 mapping['enable_metadata_optimization'],
         }
+        if port.network_type == n_constants.TYPE_VLAN:
+            mapping_dict['provider-vlan'] = True
+            mapping_dict['ext-encap-id'] = port.segmentation_id
```
rkukura, line 330: Is the physical_network stored in the endpoint file? If we really are supporting  passthru of VLAN networks, it seems it will be necessary to support a host being connected to multiple switch ports, each using a different physdom, and identified by a different physical_network name. Therefore, it seems the opflex agent will need to know which physdom this particular ext-encap-id is relative to. It is possible the opflex agent doesn't yet support multiple physdoms, so maybe it isn't needed yet.

skandasw: Thx Bob. At this point we only support a single PhysDom.

tbachman:
Thanks for the patch! Please backport through stable/queens.

PR #357, opened 2020-04-21
Set metadata and dhcp optimization to False for vlan type nets

skandasw:
The service file on opflex agent hosts triggers the creation of flows for the
optimized path. For VM's on vlan type nets we disable the optimized path.
Also added a fix to allow the co-existence a VM on a vlan type net (without
the optimized path) and a VM on a opflex type net (with the optimized path).

### DIFF
--- a/opflexagent/as_metadata_manager.py
+++ b/opflexagent/as_metadata_manager.py
@@ -316,7 +316,6 @@ class EpWatcher(FileWatcher):
     def process(self, files):
         LOG.debug("EP files: %s" % files)
 
-        curr_metadata_optimization = True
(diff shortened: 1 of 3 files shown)


skandasw:
On the original code, I am not sure if we need this:
https://github.com/noironetworks/python-opflex-agent/blob/master/opflexagent/utils/ep_managers/endpoint_file_manager.py#L176

Do we really need set the metadata optimization to True on the AAP case or let it be whatever it was ? I wrapped that around a condition for vlan nets - but wondering if we need that line even.

tbachman:
Please backport to stable/queens.

PR #360, opened 2020-04-29
Add new paramnter for nested mode

tbachman:
The opflex-agent requires a new parameter when using nested mode,
so that it knows that it should apply a datapath optimization
behavior for the access port. A new parameter is introduced for
this: access-allow-untagged. It is a boolean, with a default
value of False (i.e. if not present, implied valiue of False).

### DIFF
--- a/opflexagent/test/test_endpoint_file_manager.py
+++ b/opflexagent/test/test_endpoint_file_manager.py
@@ -317,6 +317,7 @@ class TestEndpointFileManager(base.OpflexTestBase):
                    "interface-name": 'qpi',
                    "mac": 'aa:bb:cc:00:11:22',
                    "access-interface-vlan": 4094,
+                   "access-allow-untagged": True,
                    "promiscuous-mode": mapping['promiscuous_mode'],
                    "uuid": port.vif_id + '|aa-bb-cc-00-11-22',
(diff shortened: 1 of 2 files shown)


tbachman:
The 4094 is just for a UT. It uses whatever value is provided by neutron-server.

opflexagent/test/test_endpoint_file_manager.py
```diff
@@ -317,6 +317,7 @@ def test_port_nested_domain(self):
                    "interface-name": 'qpi',
                    "mac": 'aa:bb:cc:00:11:22',
                    "access-interface-vlan": 4094,
```
kentwu1, line 319: is it ok to just assume its true when it sees the access-interface-vlan value being present?

tbachman: Unfortunately not - there are two different behaviors that we need with the access-interface-vlan: one for nested virtualization (which needs this flag present and set to True), and the other with trunks (which doesn't need this flag present, or present and set to False). The opflex-agent does different flow-mods in the two cases.

kentwu1: Got it.

PR #364, opened 2020-05-04
Handle generation of EP file for DHCP endpoint for SVI type nets

skandasw:
Since SVI type nets are not associated with an EPG and use a fake EPG
as the neutron net-id - handle the missing fields for the DHCP endpoint's
EP file generation.


tbachman:
Please backport to stable/queens.

PR #372, opened 2020-05-18
Cleanup Queens

rkukura:
Enhance compatabilty with newer Neutron branches while maintaining
compatability with stable/queens Neutron, and improve the build/test
process. Highlights include:

* Use definitions from neutron_lib where appropriate.

* Update imports of classes moved from
  neutron.agent.securitygroups_rpc to
  neutron.api.rpc.handlers.securitygroups_rpc.

* Replace RPC callback deprecated subscribe method with register.

* Don't use deprecated oslo_config.MultiConfigParser.

* Remove i18n translation markers from log messages, and update
  remaining ones to properly use oslo.i18n.

* Eliminate unneeded requirements and test-requirements, and update
  remaining ones to match upstream stable/queens Neutron.

* Use stestr instead of testr to run UTs.

* Specify basepython as python2.7 for pep8 and cover jobs, in case
  python3 version of tox is used.

* Require tox-venv to address incompatability between version 1.0 of
  MarkupSafe pinned by
  https://releases.openstack.org/constraints/upper/queens and recent
  versions of setuptools.

### DIFF
--- a/.stestr.conf
+++ b/.stestr.conf
@@ -1,0 +1,3 @@
+[DEFAULT]
+test_path=${OS_TEST_PATH:-./opflexagent/test}
+top_dir=./
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: python
 
-
 matrix:
   include:
     - python: 2.7
@@ -9,15 +8,12 @@ matrix:
     - python: 2.7
       env:
         - TOX_ENV=pep8
-    - python: 3.6
+    - python: 3.5
       env:
-        - TOX_ENV=py36
-    - python: 3.6
-      env:
-        - TOX_ENV=pep8
+        - TOX_ENV=py35
+
 install:
   - pip install tox
-
 
 script:
   - tox -e $TOX_ENV
--- a/opflexagent/_i18n.py
+++ b/opflexagent/_i18n.py
@@ -1,0 +1,20 @@
+# All Rights Reserved.
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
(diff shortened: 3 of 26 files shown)


rkukura:
> I'd be okay with merging this, provided we can verify that the change from subscribe to register for the trunk subport RPCs doesn't break the existing trunk port functionality (it seems like it shouldn't, but given that we have to release this week, I'd like to see this tested).

The two functions in stable/queens neutron are defined as follows:

def subscribe(callback, resource_type):
    # temporary hack to differentiate between callback types until the                                                                                                                                                                                      
    # 'subscribe' method is removed                                                                                                                                                                                                                         
    callback.__dict__['_ACCEPTS_CONTEXT'] = False
    _get_manager().register(callback, resource_type)


def register(callback, resource_type):
    # TODO(kevinbenton): remove this on debt collection                                                                                                                                                                                                     
    callback.__dict__['_ACCEPTS_CONTEXT'] = True
    _get_manager().register(callback, resource_type)

So subscribe and register take the same arguments and call the same function with them. The only difference is how they set _ACCEPTS_CONTEXT, which is only used in:

def push(context, resource_type, resource_list, event_type):
    """Push resource list into all registered callbacks for the event type."""

    callbacks = _get_manager().get_callbacks(resource_type)
    for callback in callbacks:
        if callback._ACCEPTS_CONTEXT:
            callback(context, resource_type, resource_list, event_type)
        else:
            # backwards compat for callback listeners that don't take                                                                                                                                                                                       
            # context and resource_type                                                                                                                                                                                                                     
            callback(resource_list, event_type)

I have added the context and resource_types args to the handle_subports method, and the UTs have verified that handle_subports is properly called with these additional args (i.e. the UT failed until I added them to handle_subports, then passed). In my opinion, this change is very low risk, but we can try adding this patch to an existing deployment to test further if needed.

skandasw:
Ran some tests on the subport related change to using register with the new method signature. Things look good Bob with some tests with Trunk port with child port add/remove etc. Details:

http://paste.openstack.org/show/793790/

PR #373, opened 2020-05-18
Add support for upstream Rocky release

rkukura:
Depend on stable/rocky rather than stable/queens branch of upstream
Neutron repository. Changes needed for compatability with stable/rocky
that were also compatible with stable/queens were made in previous
patches, so only rocky-specific changes are included here.


rkukura:
Please don't merge this until after the cleanup_queens patch has merged.

PR #386, opened 2020-06-08
Cleanup Queens (part 2)

rkukura:
Improve the build/test process. Highlights include:

* Remove the workaround for the MarkupSafe compatability issue with
  stable/queenss that required a newer version of tox.

* Fix pep8 issues that result in failures with the versions of hacking
  and flake8 used by Neutron's stable/rocky through stable/train
  branches. These changes are not necessary with the hacking and
  flake8 versions used in stable/queens, but we want to minimize code
  differences across our currently supported stable branches.

* Enable flake8-import-order and fix all the pep8 issues that it
  uncovered, particularly with order and grouping of import
  statements.

* Update pep8 configuration in tox.ini to more closely match upstream
  Neutron, and fix resulting issues. Remaining ignored checks that
  should be fixed but haven't been are marked with REVISIT in tox.ini.

### DIFF
--- a/opflexagent/test/test_async_port_manager.py
+++ b/opflexagent/test/test_async_port_manager.py
@@ -186,12 +186,12 @@ class TestAsyncPortManager(base.BaseTestCase):
 
     def test_apply_config_fails(self):
         self.agent.treat_devices_added_or_updated = mock.Mock(
-            side_effect=Exception)
+            side_effect=ValueError)
         to_schedule = set(['1', '2', '3', '4'])
         self.manager.schedule_update(to_schedule)
         update = self._get_device_requests(['1', '2', '4'])
         self.manager._opflex_endpoint_update(mock.Mock(), update)
-        self.assertRaises(Exception, self.manager.apply_config)
+        self.assertRaises(ValueError, self.manager.apply_config)
         # responses are restored after the exception
         self.assertEqual(3, len(self.manager.response_by_device_id))
 
--- a/opflexagent/utils/ep_managers/endpoint_file_manager.py
+++ b/opflexagent/utils/ep_managers/endpoint_file_manager.py
@@ -12,9 +12,9 @@
 
 import copy
 import json
-import netaddr
 import os
 
+import netaddr
 from neutron_lib import constants as n_constants
 from oslo_log import log as logging
 from oslo_serialization import jsonutils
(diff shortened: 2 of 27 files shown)


opflexagent/test/test_async_port_manager.py
```diff
@@ -191,7 +191,7 @@ def test_apply_config_fails(self):
         self.manager.schedule_update(to_schedule)
         update = self._get_device_requests(['1', '2', '4'])
         self.manager._opflex_endpoint_update(mock.Mock(), update)
-        self.assertRaises(Exception, self.manager.apply_config)
+        self.assertRaises(Exception, self.manager.apply_config)  # noqa
```
tbachman, line 194: Why is noqa needed here? Seems odd.

opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -372,8 +372,7 @@ def _mapping_to_file(self, port, mapping, fixed_ips):
         else:
             if (mapping.get('enable_dhcp_optimization', False) and
                'subnets' in mapping):
-                    self._map_dhcp_info(fixed_ips, mapping,
-                                        mapping_dict)
+                self._map_dhcp_info(fixed_ips, mapping, mapping_dict)
```
tbachman, line 375: Interesting that PEP8 doesn't flag this - I though the line above would have needed the extra 4 spaces?

opflexagent/test/test_async_port_manager.py
rkukura: Pep8 was saying that Exception is "too general", but I don't know enough about the code to know of a more specific exception to expect.

opflexagent/utils/ep_managers/endpoint_file_manager.py
rkukura: It actually complains about the extra spaces. I've had to remove them here and several other places in this repo and the GBP repo.

opflexagent/test/test_async_port_manager.py
tbachman: It's passing the exception above - you could just use something like ValueError in both places, and you could get rid of the noqa.

opflexagent/utils/ep_managers/endpoint_file_manager.py
tbachman: Ah - good to know. Thanks!

opflexagent/test/test_async_port_manager.py
rkukura: ValueError worked - thanks!

PR #389, opened 2020-06-09
Add support for upstream Stein release

rkukura:
Depend on stable/stein rather than stable/rocky branch of upstream
Neutron repository. Changes needed for compatability with stable/stein
that were also compatible with stable/rocky were made in previous
patches, so only Stein-specific changes are included here.

### DIFF
--- a/opflexagent/utils/bridge_managers/ovs_lib.py
+++ b/opflexagent/utils/bridge_managers/ovs_lib.py
@@ -13,7 +13,6 @@
 import contextlib
 
 from neutron.agent.common import ovs_lib
-from neutron.agent.ovsdb import impl_vsctl
 
 from neutron.plugins.ml2.drivers.openvswitch.agent.common import config
 from opflexagent import config  # noqa
@@ -24,11 +23,6 @@ class OVSBridge(ovs_lib.OVSBridge):
     def __init__(self, *args, **kwargs):
         self._transaction = None
         super(OVSBridge, self).__init__(*args, **kwargs)
-
-    def reset_ofversion(self):
-        context = self.ovsdb.context
-        return impl_vsctl.BaseCommand(context, 'set',
-                                      args=[self.br_name, 'protocols=[]'])
 
     @contextlib.contextmanager
     def ovsdb_transaction(self):
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -3,8 +3,9 @@
 # process, which may cause wedges in the gate later.
 hacking>=1.1.0,<1.2.0 # Apache-2.0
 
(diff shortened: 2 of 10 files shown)


rkukura:
This needs significant work before merging, primarily to deal with upstream Neutron moving in Stein from using ovs-vsctl commands to using ovsdb.

test-requirements.txt
```diff
@@ -3,8 +3,9 @@
 # process, which may cause wedges in the gate later.
 hacking>=1.1.0,<1.2.0 # Apache-2.0
 
--e git+https://opendev.org/openstack/neutron.git@stable/rocky#egg=neutron
+neutron
 
+contextlib2>=0.5.4
```
kentwu1, line 8: I'm wondering where this is from? My understanding is that this is based on this list,
https://github.com/openstack/neutron/blob/stable/stein/test-requirements.txt
but I didn't find contextlib2 there so.....

opflexagent/utils/bridge_managers/ovs_lib.py
```diff
@@ -13,7 +13,9 @@
 import contextlib
 
 from neutron.agent.common import ovs_lib
-from neutron.agent.ovsdb import impl_vsctl
+
+# REVISIT: This is gone in stein!
```
kentwu1, line 17: Is there a reason we still keep this around?

opflexagent/utils/bridge_managers/ovs_lib.py
```diff
@@ -26,9 +28,11 @@ def __init__(self, *args, **kwargs):
         super(OVSBridge, self).__init__(*args, **kwargs)
 
     def reset_ofversion(self):
-        context = self.ovsdb.context
-        return impl_vsctl.BaseCommand(context, 'set',
-                                      args=[self.br_name, 'protocols=[]'])
+        # REVISIT: impl_vsctl is gone in stein!
```
kentwu1, line 31: ditto

test-requirements.txt
tbachman: I had to add this - the python35 job was failing. I've seen references to it not being installed, even though it's considered a standard library. Adding this fixed the python35 job.

opflexagent/utils/bridge_managers/ovs_lib.py
tbachman: will fix (along with the other one I accidentally resolved).

opflexagent/utils/bridge_managers/ovs_lib.py
tbachman: Will fix.

tbachman:

PR #394, opened 2020-08-27
Add support for stable/train

tbachman:


tox.ini
```diff
@@ -14,7 +14,7 @@ usedevelop = True
 install_command =
   pip install {opts} {packages}
 deps =
-  -c{env:UPPER_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/stein}
+  -c{env:UPPER_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/train}
   -r{toxinidir}/requirements.txt
   -r{toxinidir}/test-requirements.txt
 whitelist_externals = sh
```
kentwu1, line 20: Somehow I can't comment on that specific line so just put it here. Should we use python3 instead of python2.7 on line#27 and line#32.  Neutron has moved to python 3 so,
https://github.com/openstack/neutron/blob/stable/train/tox.ini

tbachman: Yeah - we probably should. Will fix.

tbachman:
Pushed updated version, addressing reviewer comments.

PR #488, opened 2020-11-30
[HOST AGENT] Fixed lldp host-likns issue

irathore:
The compiled regex map self.port_desc_re needed to be a list for
python3.


tbachman:
Can you also do backports through stable/queens?

PR #525, opened 2021-02-15
Fix trunk subports exception

tbachman:
The RPC calls to update the trunk status were found to fail
due to concurrency issues with processing the same subports.
This patch reduces the number of RPCs needed by checking
whether subports have already been processed.

closes-bug: #1387


tbachman:
abandoning

PR #541, opened 2021-02-22
Eliminate temporary OpenStack SNAT traffic interruption after opflex agent restart

sayalinaval:
We will be using the SNAT port's MAC address as the SNAT namespace mac address. Each time the opflex agent restarts, the SNAT namespace will not be deleted and created again as long as it already exists and its mac address is the same as the mac address of the SNAT port's mac address.

### DIFF
--- a/opflexagent/test/test_endpoint_file_manager.py
+++ b/opflexagent/test/test_endpoint_file_manager.py
@@ -23,6 +23,7 @@ from opflexagent import snat_iptables_manager
 from opflexagent.test import base
 from opflexagent.utils.ep_managers import endpoint_file_manager
 
+from neutron.agent.linux import ip_lib
 from neutron.conf.agent import dhcp as dhcp_config
 from neutron_lib import constants as n_constants
 from oslo_config import cfg
@@ -43,6 +44,8 @@ class TestEndpointFileManager(base.OpflexTestBase):
         self.manager.nat_mtu_size = 9000
         self._mock_agent(self.manager)
         self.addCleanup(self._purge_endpoint_dir)
+
+        ip_lib.IPDevice = mock.Mock()
 
     def _purge_endpoint_dir(self):
         try:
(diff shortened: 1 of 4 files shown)


opflexagent/test/test_gbp_ovs_agent.py
```diff
@@ -155,7 +155,7 @@ def test_port_unbound_snat_cleanup(self):
         args_2['port'].gbp_details = mapping
```
tbachman, line 155: You may want to add:
        self.agent.ep_manager.snat_iptables.setup_snat_for_es.return_value = (
             tuple('foo-if', 'aa:bb:cc:00:11:44'))

before the next call to self.agent.try_port_binding(**args_2) on line 156. That way I think yu can keep the assertEqual to just 1.

opflexagent/test/test_endpoint_file_manager.py
```diff
@@ -44,6 +45,9 @@ def setUp(self):
         self._mock_agent(self.manager)
         self.addCleanup(self._purge_endpoint_dir)
 
+        ip_lib.IPDevice = mock.Mock()
+        ip_lib.IpLinkCommand = mock.Mock()
```
tbachman, line 49: Is the mock of IpLinkCommand  needed?

tbachman (changes requested):
Overall, very clean - just a couple of questions/comments.

opflexagent/test/test_endpoint_file_manager.py
sayalinaval: You are right, it's not needed. I will remove it.

opflexagent/test/test_gbp_ovs_agent.py
sayalinaval: I have updated the tests accordingly. Please take a look.

tbachman (approved):

tbachman:
Thanks for the patch/fix! Please post backports through stable/queens.

PR #554, opened 2021-04-06
Add a new directory in the path of notification socket.

ansao-aci:
New path is /var/run/opflex_notify/opflex-agent-notif.sock
This socket is to be shared by neutron_opflex_agent container and opflex_agent container. 
By putting the socket file in a saperate directory inside /var/run, only that directory can be conveniently shared between
the two containers as named volume.


rameescm1 (approved):

sayalinaval (approved):

tbachman (commented):
Rather than change the default, we should fix this in the automation.

PR #650, opened 2022-04-13
fix_livemigration_port_status_error

pulkitvajpayee07:


tbachman (commented):
I think you're missing a change here - in line 584 of gbp_agent.py, it restores the copies of updated and deleted objects if there was an exception. I think we need to do the same for activated_bindings:
    self.activated_bindings |= activated_bindings_copy

tbachman:
Thanks - can you squash these back into a single commit?

tbachman (approved):

tbachman:
Thanks for the patch! Please backport through stable/queens.

PR #663, opened 2022-08-02
Removal of shell calls

christides11:
Start of removing shell calls from as_metadata_manager

We want to move from calling shell calls directly as they require elevated permissions. This commit replaces shell calls related to ip commands in as_metadata_manager with equivalent calls from neutron's ip_lib, which handles permissions properly.

### DIFF
--- a/opflexagent/as_metadata_manager.py
+++ b/opflexagent/as_metadata_manager.py
@@ -26,6 +26,7 @@ import netaddr
 import pyinotify
 from six.moves import queue as Queue
 
+from neutron.agent.common import ip_lib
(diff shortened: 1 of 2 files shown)


tbachman (changes requested):
Overall, great patch! One note on the commit message: the convention is to limit all lines to less than 80 characters (except the title, which I think is supposed to be 60 or less?). Other than that, LGTM!

sayalinaval (approved):

irathore (approved):

tbachman (approved):

tbachman:
Please create backports through stable/queens.

PR #665, opened 2022-08-23
Replaced shell calls in as_metadata_manager

christides11:
Replaced sh calls with the execute method from neutron's agent utils.


tbachman (approved):

tbachman:
Thanks for the patch! Please provide backports through stable/queens.

PR #793, opened 2022-12-06
Restart FileWatcher child process

christides11:
FileWatcher will restart the FileProcessor child process when needed.


tbachman (commented):
A couple of questions:
1) If you stop the entire ciscoaci_neutron_opflex_agent container, does the child process get terminated?
2) Is this patch meant to address the case where the parent process crashes?

tbachman:
Please backport through stable/queens. Thanks for the patch!

PR #827, opened 2023-02-27
Fix destination ip mapping

christides11:
Makes AsMetadataManager not delete the anycast_services file.


tbachman (commented):
Looks good - and this has been tested on a deployed setup?

tbachman (approved):

PR #828, opened 2023-03-08
Reuse SNAT floating IPs when Available

christides11:
Checks EP files for SNAT floating IPs and reuse them if we can.


tbachman:
Please provide backports through stable/queens.

PR #851, opened 2023-04-28
Fixed has_ip check in AsMetadataManager

christides11:
Fixed the text match case so that it checks for the proper ip.


tbachman (approved):

tbachman:
Thanks for the patch! Please provide backports through stable/queens.

PR #868, opened 2023-06-09
Remove python2 (it is dead)

irathore:
Though we are removing python2 using ``sed``. ``build-rpm.sh`` still had python2 which does not exist anymore.


tbachman:
Please squash commits

PR #948, opened 2024-02-29
Add Bobcat Support

nisar123456:


opflexagent/test/test_opflex_notify.py
```diff
@@ -32,6 +35,19 @@ def setUp(self):
...
+                               f"{sys.version_info.minor}"
+        python_version_long = f"python{sys.version_info.major}." \
+                              f"{sys.version_info.minor}"
+        self.policy_d_path = os.path.join('.tox',
+                python_version_short, 'lib', python_version_long,
+                'site-packages', 'neutron', 'tests', 'etc', 'policy.d')
+
+        os.makedirs(self.policy_d_path, exist_ok=True)
+
+    def tearDown(self):
```
tbachman, line 48: Just curious - why was this needed?

nisar123456: testcase are expecting an empty dictionary in the test environment, which is why we used this to create it.

tbachman (approved):

PR #949, opened 2024-05-17
Fixing Static Analysis Issues

christides11:
Addresses issues found when running the bandit static analysis tool.

### DIFF
--- a/opflexagent/as_metadata_manager.py
+++ b/opflexagent/as_metadata_manager.py
@@ -17,8 +17,9 @@ import multiprocessing
 import os
 import os.path
 import signal
-import subprocess
+import subprocess  # nosec
 import sys
+import tempfile
 import time
 import uuid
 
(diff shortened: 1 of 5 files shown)


sbagalur (commented):
for the try except pass error, can u add a comment inside that block instead of nosec.

opflexagent/as_metadata_manager.py
sbagalur, line 1: Since the code for the exception part is not an error, might want to use LOG.info. Also can remove the pass.

opflexagent/opflex_conn_track.py
sbagalur, line 1: can remove the 'pass' and change log levels

opflexagent/vpplib/vpp_papi_provider.py
sbagalur, line 1: same as above comment

opflexagent/as_metadata_manager.py
sbagalur, line 1: for the temp directory, maybe this is a possibility to fix the issue. Or you might find other functions in the tempfile library without static analysis errors.  https://security.openstack.org/guidelines/dg_using-temporary-files-securely.html
import tempfile

filedir = tempfile.gettempdir()
Test and confirm.

christides11: Tested and it seems to have gotten rid of the issue.

sbagalur (approved):

tbachman (approved):

PR #1003, opened 2025-03-02
Remove Neutron Server From Debian

nisar123456:


tbachman (changes requested):
Can we instead do this against master and have backports?

PR #1117, opened 2026-05-26
Add support for Distributed SNAT

nisar123456:
The agent consumes distributed SNAT information from gbp RPCs,
creates SNAT and service configuration files for opflex-agent,
tracks local and remote service node mappings, updates endpoint
SNAT references, and reports allocated SNAT IP/port mappings
through the dist_snat_mappings agent configuration field.

### DIFF
--- a/opflexagent/utils/ep_managers/endpoint_file_manager.py
+++ b/opflexagent/utils/ep_managers/endpoint_file_manager.py
@@ -24,6 +24,7 @@ from oslo_utils import uuidutils
 
 from opflexagent import as_metadata_manager
 from opflexagent import constants as ofcst
+from opflexagent import distributed_snat_manager
 from opflexagent import snat_iptables_manager
 from opflexagent.utils.ep_managers import endpoint_manager_base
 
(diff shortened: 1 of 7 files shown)


opflexagent/utils/ep_managers/endpoint_file_manager.py
```diff
@@ -562,6 +564,13 @@ def filter_cidr_aaps(aaps):
             if self.disable_security_group_for_nested_mode:
                 mapping_dict.pop('security-group', None)
 
+        # Add distributed SNAT UUID references from host_snat_ips in the RPC.
+        dist_snat_entries = self.dist_snat_manager.build_dist_snat_entries(
+            mapping, mapping_dict)
+        self.dist_snat_manager.sync_endpoint(mapping_dict['uuid'],
+                             dist_snat_entries,
+                             mapping_dict)
```
tbachman: Would this create SNAT configuration for both distributed SNAT and the existing SNAT? There is a call above to _handle_host_snat_ip. I'm wondering if this code should instead be moved into that method, where one or the other is handled for a given SNAT IP.

tbachman (changes requested):
One question to follow up on. Otherwise, looks good!

